PR #2818, opened 2018-01-21
Port build to Dune and require 4.06.0

avsm:
This mega-pr does the following:

- fix ocaml-topexpect to compile with 4.06.0
- bump the minimum requirement to build to 4.06.0
- port the build to use Dune:
  * use `make dep` to regenerate the jbuild files
  * use `make` to build the site
  * use `make code` to check the topexpect files and `jbuilder promote` to update any in-tree
- add various dummy `.ml` files to suppress Dune warnings when we didn't have a corresponding `.mli` file.

There is still some work to be done before this PR can be merged:
- [x] need to extend the `examples/` jbuild support to include shell script dependencies.  I plan to do this via ` jbuild.deps` sexpression fragment that is parsed by `rwo-jbuild` to add the dependencies.
- [x] fix up the failing shell script runs in examples/ which have regressed due to the above.
- [x] rewrite `jbuild` files in published examples to remove the `(include jbuild.inc)` that is necessary for the "host" build of the book
- [x] pygments is currently disabled, but we can perhaps restore this via highlight.js instead of statically running it. Can probably merge without this.


avsm:
pushed support for `jbuild.deps` in a1fcec103abd5fcdeeb75c48e6d50c84b1b6dc6e -- now need to port all the failing examples.

Some things are still failing due to the `jbuild.inc` issue (its not being copied into the _build directory)

avsm:
I've also fixed the .sh execution to avoid capturing ANSI characters, and significantly improved the CLI/JSON chapters

avsm:
i think this is good enough to merge now.  Will check that the site rebuilds next after the merge.

PR #2821, opened 2018-01-23
Base in prologue

yminsky:
Switched the prologue to focus on Base rather than Core_kernel


book/00-prologue.html
```diff
@@ -182,67 +182,81 @@ <h2>A Brief History</h2>
...
+
+        <p>However good it is, a language on its own isn't enough. You
+          also need a set of libraries to build your applications
+          on. A common source of frustration for those learning OCaml
+          is that the standard library that ships with the compiler is
+          limited, covering only a subset of the functionality you
+          would expect from a general-purpose standard library. That's
+          because the standard library isn't really a general-purpose
+          tool; its fundamental role is in bootstrapping the compiler,
+          and has been purposefully kept small and simple.</p>
```
avsm: perhaps small and portable?

book/00-prologue.html
```diff
@@ -182,67 +182,81 @@ <h2>A Brief History</h2>
...
-        systems. All of these libraries are distributed under a
-        liberal Apache 2 license to permit free use in hobby,
-        academic, and commercial settings.</p>
+          more than 15 years, developed the code in <code>Base</code>
+          for its own internal use, but from the start designed it
+          with an eye toward being a general-purpose standard
+          library. Like the OCaml language itself, <code>Base</code>
+          is engineered with correctness, reliability, and performance
+          in mind.  It's also designed to be easy to install and
+          highly portable, available on every platform OCaml is,
```
avsm: this `, available on every platform OCaml is,` reads oddly

book/00-prologue.html
yminsky: Accepted.

book/00-prologue.html
yminsky: Good catch. Fixed.

avsm:
A more general comment for later is that we need to figure out how to render lower case opam in the book so it doesnt look odd. upstream decided that it's called opam and not Opam or OPAM...

PR #2829, opened 2018-01-23
fix sexpr example dependencies

avsm:


avsm:
I'll document the deps in the README, still need to refresh it.

PR #2831, opened 2018-01-23
ffi build and API fixes

avsm:


examples/code/ffi/posix.topscript
```diff
@@ -82,12 +83,6 @@ let gettimeofday' () =
...
-Characters 59-71:
-Error: Unbound value gettimeofday
-|}];;
+[%%expect ocaml {|val gettimeofday' : unit -> float = <fun>|}];;
 gettimeofday' () ;;
-[%%expect{|
-Characters 0-13:
-Error: Unbound value gettimeofday'
-|}];;
+[%%expect ocaml {|- : float = 1516746708.884176|}];;
```
yminsky: Note that this is non-deterministic. Maybe later it would make sense to wrap a different system call whose output we can arrange to be deterministic?

avsm: Yeah, I've made a note of this when doing my Ctypes refresh

PR #2836, opened 2018-01-25
travis: simplify to use the non-Docker build

avsm:
This requires manually updating the depexts, but will hopefully
be more reliable on Travis using the 'native' build infrastructure.

Downside is that we need to manually encode the depexts in the
Travis file and keep it updated by hand.


avsm:
I've got a simpler solution for this soon with Buildkite and the new cluster

PR #2837, opened 2018-01-27
Base in lists and patterns

yminsky:
This is a rework of the chapter to get it set up to use Base. The biggest change comes from the fact that polymorphic compare is not enabled by default, which leads to the addition of a rather long aside about Base.Poly.


book/03-lists-and-patterns.html
```diff
@@ -463,45 +466,60 @@ <h3>Partitioning with List.partition_tf</h3>
...
-          <p>Another very common operation on lists is
-          concatenation. The list module actually comes with a few
-          different ways of doing this. First, there's
-          <code>List.append</code>, for concatenating a pair of
-          lists:<idx>lists/combining</idx><idx>List
-          module/List.append</idx></p>
-          <link rel="import" href="code/lists-and-patterns/main.topscript" part="32"/>
+          <p>
+	    Another very common operation on lists is
+            concatenation. The list module actually comes with a few
```
avsm: s/list module/List module

book/03-lists-and-patterns.html
```diff
@@ -634,64 +654,113 @@ <h1>Terser and Faster Patterns</h1>
...
+        <p>
+	  You might wonder how you could build functions like these
+          yourself if OCaml didn't come with them built in. It turns
+          out that you <em>can't</em> build these functions on your
+          own. OCaml's polymorphic comparison functions are built into
+          the runtime to a low level. These comparisons are
+          polymorphic on the basis of ignoring almost everything about
+          the types of the values that are being compared, paying
+          attention only to the structure of the values as they're
+          laid out in memory.</p>
```
avsm: Perhaps forward reference to Part III and the runtime memory layout section here.

PR #2838, opened 2018-01-27
Base in variables and functions

yminsky:
This one was pretty easy. Still have some references to Core and Stdio, but Base is now opened in place of Core_kernel.


yminsky:
Hmm. Looking at the diffs, something is wrong here. This is recapitulating some of the changes in the lists chapter. I'll figure it out.

PR #2840, opened 2018-01-27
Base in fmap

yminsky:
Still having some issues here. The generated book, in particular still contains references to core_kernel. Is some of the generation of output not yet hooked up?


yminsky:
This is from the generated HTML, which highlights the problem.

image

yminsky:
The example now looks better:

image

in that it now refers to Base being missing. It's not great to have all of those color escape codes in there. We either need to interpret them or suppress them.

avsm:
I'm going to merge this and look at the linkage issues in trunk

PR #2847, opened 2018-01-28
Base in error handling

yminsky:
Pretty simple; had to open Stdio in a few places, and some other minor tweaks.

### DIFF
--- a/book/07-error-handling.html
+++ b/book/07-error-handling.html
@@ -184,6 +184,8 @@
             is distributed with Base and is the most common
             serialization format used in Base. Indeed, most types in
             Base come with built-in s-expression converters.
+
+
             <idx>Sexplib package/sexp converter</idx></p>
           <link rel="import" href="code/error-handling/main.topscript" part="7"/>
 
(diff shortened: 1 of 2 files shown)


book/07-error-handling.html
```diff
@@ -181,9 +181,9 @@ <h2>Error and Or_error</h2>
 
           <p>
 	    S-expressions are supported by the Sexplib package that
-            is distributed with Core and is the most common
-            serialization format used in Core. Indeed, most types in
-            Core come with built-in s-expression converters.
+            is distributed with Base and is the most common
```
avsm: forward reference to Part II and the serialization chapter?

book/07-error-handling.html
```diff
@@ -446,7 +446,7 @@ <h1>Declaring Exceptions Using <code>[@@deriving sexp]</code></h1>
           <h2>Helper Functions for Throwing Exceptions</h2>
 
           <p>
-	    OCaml and Core provide a number of helper functions to
+	    OCaml and Base provide a number of helper functions to
```
avsm: Just a nit here, but is "OCaml and Base" necessary, as opposed to just "Base"?  Anything that OCaml provides is re-exported by Base, often with different conventions, so adding OCaml here seems redundant.

book/07-error-handling.html
yminsky: There's already a forward reference a few lines down.

book/07-error-handling.html
yminsky: Fixed.

PR #2848, opened 2018-01-28
Base in records

yminsky:
Pretty simple. I deleted some work I'd done to explain all the various and sundry OCaml string syntaxes, because it wasn't necessary here. We'll probably need it somewhere, eventually.

### DIFF
--- a/book/05-records.html
+++ b/book/05-records.html
@@ -37,8 +37,8 @@
       Here's a simple example: a <code>service_info</code> record that
       represents an entry from the <code>/etc/services</code> file on
       a typical Unix system. That file is used for keeping track of
-      the well-known port and protocol name for protocols such as ftp
-      or ssh.  Note that we're going to open <code>Core</code> in this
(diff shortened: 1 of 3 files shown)


book/05-records.html
```diff
@@ -33,11 +33,15 @@ <h1>Records</h1>
...
-    that represents an entry from the <code>/etc/services</code>
-    file on a typical Unix system. That file is used for keeping
-    track of the well-known port and protocol name for protocols
-    such as ftp or ssh.</p>
+    <p>
+      Here's a simple example: a <code>service_info</code> record that
+      represents an entry from the <code>/etc/services</code> file on
+      a typical Unix system. That file is used for keeping track of
+      the well-known port and protocol name for protocols such as ftp
+      or ssh.  Note that we're going to open <code>Core</code> in this
```
avsm: I think ftp and ssh are usually capitalised as FTP and SSH, since they are acronyms

yminsky: Will do.

PR #2849, opened 2018-01-28
Base in imperative programming

yminsky:
Lots of small changes here, much of which was fixing breakage that was already committed.


avsm:
lgtm

PR #2850, opened 2018-01-31
Update paths for vim configuration

nbuonin:
I ran into two issues while setting up Vim.  All Merlin commands would return "Not an editor command", and ocp-indent would throw errors when opening Vim, saying `the file substitute(system('opam config var share'), '\n$', '', '''') . "/typerex/ocp-indent/ocp-indent.vim" can not be opened`.

Greping through ~/.opam found the updated paths.  Appending the new ocp-indent path to Vim's run time path got it working.


avsm:
thanks!

PR #2854, opened 2018-02-20
Update 07-error-handling.html

constfun:
V1 example serialized a "time", V2 example serializes a char, looks like a typo to me.

### DIFF
--- a/book/07-error-handling.html
+++ b/book/07-error-handling.html
@@ -190,7 +190,7 @@
           <link rel="import" href="code/error-handling/main.topscript" part="7"/>
 
           <p>
-	    Note that the time isn't actually serialized into an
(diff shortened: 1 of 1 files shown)


book/07-error-handling.html
```diff
@@ -190,7 +190,7 @@ <h2>Error and Or_error</h2>
           <link rel="import" href="code/error-handling/main.topscript" part="7"/>
 
           <p>
-	    Note that the time isn't actually serialized into an
```
yminsky: Good catch, but maybe better, "Note that the character isn't actually..." rather than the more generic "type"?

constfun: "character" is definitely better, I had the same thoughts shortly after submitting.

PR #2856, opened 2018-02-28
Remove dead code

samoht:
This is built on top of #2855

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,14 +1,14 @@
 .PHONY: all clean dep code publish
 
 all:
-	@jbuilder build @site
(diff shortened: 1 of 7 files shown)


avsm:
@samoht note that the non-deterministic flag to topexpect still needs to be plumbed into the gen_jbuild rules somehow, for the topexpect files that are nondeterministic

yminsky:
Just to clarify, topexpect files are not non-deterministic or deterministic as a whole: individual clauses within a topexepect file should be individually flaggable as non-deterministic. So the ability to run the non-deterministic tests should just be available everywhere, I think.

PR #2858, opened 2018-03-02
Remove {raw}.sexp files

samoht:
The interesting commit here is 1b3a1c7 where the raw `*.sexp` files are removed as these files are now read directly by the HTML generator. (This is based on top of #2855, #2856 and #2857).

The last commit (a075abc) update the auto-generated files.


samoht:
Updated.

yminsky:
@avsm should probably review this change, since I think he understands the code a good bit better.

avsm:
lgtm

PR #2859, opened 2018-03-02
Remove stamp files

samoht:
(This is based on top of #2855, #2856 and #2857).

The interesting commit here is fccb947 which replaces stamp files with a `code` alias.  9ff150d shows the difference in the auto-generated files, after running `make dep`.


samoht:
Updated (and sorry for the unrelated branch name)

PR #2861, opened 2018-03-02
Add support for cram tests

samoht:
This is based on #2858, #2859 and #2860.

That PR first imports the [cram](https://github.com/ocaml/dune/blob/master/test/blackbox-tests/cram.mll) executable from `dune` and then generates the rules in 54a7c51 to use them in rwo. 613e79a shows how the `build.inc` rules are modified.

This make all the existing `.errsh` and `.sh`scripts obsolete so some work will need to be done to update their contents.


bin/lib/#rwo_scripts.ml#
```diff
@@ -0,0 +1,255 @@
+open Core
```
yminsky: This file is a mistake, yes? Just leftover editor cruft?

samoht:
I have cleaned up the PR so it is easier to review now. But I don't think that PR is ready to review just yet as I need to update the `.sh` an `.errsh` scripts to be compatible with the cram format.

yminsky:
Ok. Happy to wait.

samoht:
This is ready to review. The patch is quite big: the first part is the addition of the tool for doing cram tests; the remaining commits are removal of most of the `jbuild.deps` files and then updating the shell scripts to the new cram format.

Some tests are failing for various reasons. I am not sure this PR is the right place to fix these, as it is already quite large... Let me know what you prefer.

examples/code/async/echo/run_echo.sh
```diff
@@ -1,6 +1,17 @@
...
+  [1mFile "[1mecho.ml", line 11, characters 19-25[0m[0m:
```
yminsky: The escape codes in here look a little dicey. Is there something we can do about that?

examples/code/async/search_with_error_handling/run_search_with_error_handling.sh
```diff
@@ -1,2 +1,16 @@
-jbuilder build search_with_error_handling.exe
-./_build/default/search_with_error_handling.exe -servers localhost,api.duckduckgo.com "Concurrent Programming" OCaml
+  $ jbuilder build search_with_error_handling.exe
+  $ ./_build/default/search_with_error_handling.exe -servers localhost,api.duckduckgo.com "Concurrent Programming" OCaml
+  Concurrent Programming
+  ----------------------
+  
+  DuckDuckGo query failed: Unexpected failure
```
yminsky: Can we do something to mark these broken examples so we can go back to them and fix them later, maybe?

examples/code/async/echo/run_echo.sh
```diff
@@ -1,6 +1,17 @@
...
+  [1;31mError[0m: This expression has type Core.Bytes.t = bytes
+         but an expression was expected of type string
+  [1]
+  $ ./_build/default/echo.exe &
+  $ sleep 1
+  sh: ./_build/default/echo.exe: No such file or directory
+  $ echo "This is an echo server" | nc 127.0.0.1 8765
```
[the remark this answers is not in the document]
samoht: because of the previous error message:
```
This expression has type Core.Bytes.t = bytes
         but an expression was expected of type string
```

examples/code/async/echo/run_echo.sh
samoht: Not sure if we can turn off colours with jbuilder (apparently not).

PR #2863, opened 2018-03-02
optimize the Dockerfile

samoht:
Start by installing everything before copying the source tree: this
way the dependencies can be cached and do not have to be reinstalled
everytime a source file is changed.


samoht:
So with the new aggressive caching strategy the PR completes in ~5min instead of ~15min. That looks like a good improvement vs. the loose in flexibility/automation (as one needs to list explicitly all the dependency already presents in the Dockerfile).

PR #2871, opened 2018-03-07
Update toplevel scripts

samoht:
4.06 has changed the way weak polymorphic variables are printed. Also, there are a few breakage due to the switch to Base, which can be fixed by following commits.

My plan is to fix the broken toplevel scripts with new, more targeted PRs later on.

### DIFF
--- a/examples/code/async/main.mlt
+++ b/examples/code/async/main.mlt
@@ -138,15 +138,15 @@ let count_lines filename =
 [@@@part "15"];;
 let ivar = Ivar.create ();;
 [%%expect ocaml {|
-val ivar : '_a Ivar.t =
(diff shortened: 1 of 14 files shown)


examples/code/imperative-programming/file2.topscript
```diff
@@ -14,23 +14,38 @@ let sum_file filename =
     List.fold ~init:0 ~f:(+) numbers)
     ~finally:(fun () -> In_channel.close file)
 ;;
```
yminsky: Should we maybe just add open Stdio at the top of this? We could do it in a subfeature, but it's such an easy fix, and this way we can't forget to fix it...

PR #2873, opened 2018-03-08
Fix GADT example

samoht:


samoht:
I am merging this quickly as it is breaking `make code`. The only "fix" in the code was to add compatibility with `bytes` by using `Bytes.of_string "foo"` instead of "foo".

PR #2881, opened 2018-03-11
Base in maps v2

yminsky:
This is a pretty big rewrite, and reading the diff on github isn't easy, especially because the text and the mlt code aren't together in one file. If you want to read it over, probably the easiest way is to generate the HTML and read it.

There's more I should maybe add in a followup patch, including a section describing how to use Map.M(Int).t and why it's useful.

Also, there's an unrelated fix to an example in the first-class-module section where a section was marked as non-deterministic. This was getting in the way of the build.

### DIFF
--- a/examples/code/maps-and-hash-tables/main.mlt
+++ b/examples/code/maps-and-hash-tables/main.mlt
@@ -260,21 +260,14 @@ Error: This expression has type (int, string, Int.comparator_witness) Map.t
 |}];;
 
 [@@@part "17"];;
-let dedup ~comparator l =
-  List.fold l ~init:(Set.empty ~comparator) ~f:Set.add
-  |> Set.to_list
-;;
-[%%expect{|
-Characters 58-68:
-Error: The function applied to this argument has type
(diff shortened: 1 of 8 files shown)


examples/code/maps-and-hash-tables/main.mlt
```diff
@@ -30,42 +30,127 @@ List.Assoc.add ~equal:Int.equal digit_alist 0 "zilch";;
...
+
+[@@@part "mc.2.2"];;
+module Book = struct
+
+  type t = { title: string; isbn: string }
+
+  let compare t1 t2 =
+    let cmp_title = String.compare t1.title t2.title in
+    if cmp_title <> 0 then cmp_title
+    else String.compare t1.isbn t2.isbn
```
avsm: Just a minor stylistic point here; would this be clearer as a pattern match?

```
let compare t1 t2 =
  match String.compare t1.title t2.title with
  | 0 -> String.compare t1.isbn t2.isbn
  | n -> n
```

PR #2882, opened 2018-03-11
Needed to add the flag -no-alias-deps to get the example to build.

linse:
I'm completely new to this and would love to learn more about this flag. :smiley_cat: 
Thx @stedolan for helping me debug!


yminsky:
What compiler version are you running this on?  I'm currently running on 4.06.0, and don't see any problem with the script as in the book.

stedolan:
I'm not sure what version @linse's on, but I saw the same problem on 4.06.0 (using base v0.9.3, stdio v0.9.0).

stedolan:
@yomimono points me towards ocaml/dune#567, which does look like the same issue. The error message I'm seeing is:

```
File "freq.ml", line 1:
Error: No implementations provided for the following modules:
         Stdio__ referenced from freq.cmx
         Base__ referenced from freq.cmx
```

yminsky:
Closing. It sounds like this is a bug in a jbuilder beta that will be resolved soon, so it doesn't seem sensible to change RWO for this purpose.

PR #2883, opened 2018-03-12
Add non-deterministic commands in cram tests

samoht:
To add non-deterministic test commands, use:

```
   $ <command>
@@ --non-deterministic
  <output>
  ...
@@ exit <exit code>
```

As previously, the exit code is printed only when it is non-zero. But the pretty-printer now has an optional `hide` argument which controls if lines starting by `@@` printed or not. The HTML generator uses this to hide `--non-deterministic` and `exit` lines in the HTML output.


samoht:
Rebased and updated; There is now 2 modes of non-determinism:

### Non-deterministic outputs

You want to run the tests (because it generates some artefacts that you might need later for instance) but you don't want to check that the outputs match.

```
   $ <command>
@@ --non-deterministic
  <output>
  ...
@@ exit <exit code>
```

### Non-deterministic commands

You don't even want to run the command (as it is a benchmarks which takes a very long time):

```
@@ --non-deterministic
   $ <command>
  <output>
  ...
@@ exit <exit code>
```

samoht:
I've updated the design, where I have separated more clearly pre and post conditions; I am not totally happy with the syntax but that's easy enough to change later it's it too heavy in practice.

examples/code/gc/barrier_bench/barrier_bench.sh
```diff
@@ -0,0 +1,110 @@
...
+%% --non-deterministic [skip]
+  $ ./_build/default/barrier_bench.exe -ascii alloc -quota 1
+  Estimated testing time 2s (2 benchmarks x 1s). Change using -quota SECS.
+
+    Name        Time/Run   mWd/Run   mjWd/Run   Prom/Run   Percentage
+   ----------- ---------- --------- ---------- ---------- ------------
+    mutable       8.98ms    2.00Mw     20.36w     20.36w      100.00%
+    immutable     5.66ms    5.00Mw                             63.08%
+
+### help
```
yminsky: The help generation here seems awfully big. Probably an ellipsis is desired, rather than printing out the entire help text, no?

samoht: As far as I know, there is no first-class support for ellipsis. But something like adding `...` and non-deterministic annotations would work -- is it what you had in mind?

yminsky:
@let-def actually added first-class support for ellipses. If you put ... on a line on its own, then that alters the matching function to consider it equal to the generated output as long as it can be matched in the obvious way.

PR #2884, opened 2018-03-14
More base conversions

yminsky:
Move the first-class-modules and functors chapters to Base, mostly.

### DIFF
--- a/examples/code/fcm/query_handler.mlt
+++ b/examples/code/fcm/query_handler.mlt
@@ -4,7 +4,7 @@
 #require "ppx_jane";;
 #install_printer Core.Sexp.pp_hum;;
 #install_printer Core.Error.pp;;
-open Core_kernel;;
+open Base;;
 [@@@part "0.5"];;
 module type Query_handler = sig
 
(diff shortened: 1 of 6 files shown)


examples/code/fcm/query_handler.mlt
```diff
@@ -52,7 +52,7 @@ val sexp_of_u : u -> Sexp.t = <fun>
 |}];;
 sexp_of_u {a=3;b=7.};;
 [%%expect ocaml {|- : Sexp.t = ((a 3) (b 7))|}];;
-u_of_sexp (Sexp.of_string "((a 43) (b 3.4))");;
+u_of_sexp (Core_kernel.Sexp.of_string "((a 43) (b 3.4))");;
```
avsm: Why is Core_kernel still needed here?  I thought Sexp was part of Base...

examples/code/fcm/query_handler.mlt
```diff
@@ -206,13 +206,16 @@ val dispatch :
   Sexp.t -> Sexp.t Or_error.t = <fun>
 |}];;
 [@@@part "13"];;
+open Stdio;;
 let rec cli dispatch_table =
   printf ">>> %!";
   let result =
-    match In_channel.input_line stdin with
+    match In_channel.input_line In_channel.stdin with
```
avsm: `match In_channel.(input_line stdin)` instead?

examples/code/fcm/query_handler.mlt
yminsky: Base has s-expressions, but no printers or parsers. Those are in Parsexp and Sexplib. Core_kernel does include those. Maybe we should reference Parsexp directly, but the API is a little more complicated.

examples/code/fcm/query_handler.mlt
yminsky: A good idea. I'll do that.

PR #2888, opened 2018-03-18
Using maps in deriving

yminsky:
Add another small section to the maps chapter, covering how you use [@@deriving] on map or set types.

### DIFF
--- a/book/13-maps-and-hashtables.html
+++ b/book/13-maps-and-hashtables.html
@@ -482,6 +482,54 @@
           </aside>
 	</section>
 
+	<section id="applying-deriving-to-maps-and-sets" data-type="sect2">
+	  <h2>Applying <code>[@@deriving]</code> to maps and sets</h2>
+
+	  <p>
+	    In the previous section, we showed how to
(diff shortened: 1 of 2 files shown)


book/13-maps-and-hashtables.html
```diff
@@ -481,6 +482,54 @@ <h5>=, ==, and phys_equal</h5>
...
+	    can see below.</p>
+
+	  <link rel="import" part="adm.2"
+		href="code/maps-and-hash-tables/main.mlt" />
+
+	  <p>
+	    Here, we use a functor, <code>Map.M</code>, to define the
+	    type we need. While this looks different than the ordinary
+	    type signature, the meaning of the type is the same, as
+	    evidenced here.</p>
```
avsm: The grammatical structure of these two paragraphs seems to be a little verbose:

```
There is, however, ..., which you can see below.
 ...
 Here, ..., as evidenced here.
``` 

Consider compressing some of that?

yminsky: I tweaked the language a bit to make it flow better.

PR #2890, opened 2018-03-19
Remove jbuild.deps files

samoht:
.mlt files now implicitely depends on all files in the same directory and
its subdirectories.


avsm:
Yes this is simpler and matches jbuilder's own semantics for source files

PR #2894, opened 2018-03-19
topexpect: add support for the #show directives

samoht:


yminsky:
What is #show? Does this relate to #typeof, which is currently used in the book?

https://github.com/realworldocaml/book/blob/d60a993bfd6b35b1fae8f609d0cef363a6eaacf6/examples/code/maps-and-hash-tables/main.mlt#L47

samoht:
`#show` is the directive used in the usual OCaml toplevel. I am not sure what `#typeof` is but I guess it is the same thing :-) Where is this defined and what it is supposed to do?

yminsky:
In utop, there's a #typeof directive, but no #show. We're encouraging people to use utop, so my tendency is to think we should match utop's toplevel directives. It is odd that they're different, though.

avsm:
I think `#show` might be more modern -- it didn't exist before 4.04.0, whereas utop's `#typeof` came earlier.  The main difference seems to be that #show will follow module aliases and also doesnt require quotes around the type.

https://github.com/ocaml/ocaml/pull/520

We should probably stick to the standard toplevel where possible.  Should we consider deprecating #typeof in utop if there is no pressing need for an alternative syntax @diml @rgrinberg?

yminsky:
Sounds fine to me. Right now, #show doesn't work at all in stop, so we'll need to add it.

yminsky:
Never mind. #show works fine on utop. I remove all objections.

PR #2898, opened 2018-03-20
Fix HTML typos

samoht:


samoht:
Merging this quickly as it breaks the build.

Issue #2907, opened 2018-03-22
HTML generation is messed up


yminsky:
A few examples from what's currently up on dev.realworldocaml.org

![selection_001](https://user-images.githubusercontent.com/715302/37744957-7d8a4fda-2d48-11e8-940c-f6bb296fd35e.png)
![selection_002](https://user-images.githubusercontent.com/715302/37744959-817ea00a-2d48-11e8-9a70-3412b988d3d9.png)

This seems like a recent regression.

samoht:
This is probably introduced by #2897 I need to check why entities are not resolved properly with lambdasoup.

samoht:
This is now fixed; #2910 will fix the other issue where words are being a bit shuffled...

PR #2909, opened 2018-03-22
Remove HTML encoding -- lambdasoup is handling this for us now

samoht:


samoht:
Should fix #2907

samoht:
Merging quickly to fix the rendering of https://dev.realworldocaml.org

PR #2910, opened 2018-03-22
Fix another HTML rendering bug

### LINKED ISSUE #2906: Weird problems with text
Under macOS (Chrome or Safari), I'm getting things like this (from install.html):

image

Clearly that should read `To install OCaml from source, you'll first need a C compiler, generally gcc or llvm.`

These problems are pervasive throughout — whenever anything is displaying text in multiple fonts, some of the text will end up at the end of the paragraph. Again, I'm getting this both under Chrome and Safari. I can supply lots more screenshots.

samoht:
Fix #2906


samoht:
Merging quickly to improve rendering on dev.realocaml.org

PR #2917, opened 2018-03-23
Update to v0.11

yminsky:
This updates to the v0.11 version of Base and other JS libraries. Also it adds some non-deterministic markers.


avsm:
Looks like another change is needed for rwo_toc:

```
File "bin/lib/rwo_toc.ml", line 202, characters 21-63:
Error: The function applied to this argument has type
         cmp:('a -> 'a -> int) -> 'a list -> 'a list
This argument cannot be applied with label ~compare
```

examples/code/command-line-parsing/basic_md5_with_optional_file_broken/build_basic_md5_with_optional_file_broken.errsh
```diff
@@ -1,8 +1,8 @@
...
   File "basic_md5_with_optional_file_broken.ml", line 17, characters 4-11:
-  Error: This expression has type string -> unit -> unit
+  Error: This expression has type Base.string -> unit -> unit
```
avsm: It's a pity short paths doesnt convert this Base.string to string

examples/code/error-handling/exn_cost/run_exn_cost.sh
```diff
@@ -1,39 +1,11 @@
...
-  File "exn_cost.ml", line 10, characters 33-37:
-  Warning 3: deprecated: Exit
-  [2016-09] this element comes from the stdlib distributed with OCaml.
-  Refering to the stdlib directly is discouraged by Base. You should either
```
avsm: I notice this typo is still present in the latest Base (`refering` -> `referring`)

examples/code/command-line-parsing/basic_md5_with_optional_file_broken/build_basic_md5_with_optional_file_broken.errsh
yminsky: When Leo's short-paths patch lands, this issue should be resolved.

yminsky:
Maybe you were on the wrong version when building? rwo_toc.ml is already fixed in this feature, and I verified that it builds.

samoht:
CI cache needs to be updated manually whenever the dependencies are changing by doing:

```
make docker
docker push ocaml/rwo
```

samoht:
Note this breaks the compilation of examples, as `async` examples are using `cohttp-async` which do not compile with base v0.11.0 yet.

PR #2904, opened 2018-03-21
Tool to convert from HTML to markdown

samoht:
Tracking issue is https://github.com/realworldocaml/book/issues/2929

PR #2928, opened 2018-03-26
Convert all chapters to markdown

samoht:
This is build using `html2md` (see #2904) with a few tweaks to the HTML contents (#2927).

The process is reproducible and lossless when using with something like:

```
pandoc --section-divs -f markdown-smart-auto_identifiers -t html5 chapter.md -o chapter.html 
```

The result HTML are different, but `html2md` will reconstruct the same AST for the initial and new HTML files.


book/18-concurrent-programming.md
```diff
@@ -0,0 +1,1062 @@
...
+that may require simultaneous waiting is given an operating system thread of
+its own so it can block without stopping the entire program.
+<idx>threads/preemptive vs. single-threaded programs</idx>
+
+ Another approach is to have a single-threaded program, where that single
+thread runs an *event loop* whose job is to react to external events like
+timeouts or mouse clicks by invoking a callback function that has been
+registered for that purpose. This approach shows up in languages like
+JavaScript that have single-threaded runtimes, as well as in many GUI
+toolkits. <idx>event loops</idx><idx>system threads</idx>
```
yminsky: It might be nice to have nicer way of expressing index terms (but it's not a big deal.)

book/18-concurrent-programming.md
```diff
@@ -0,0 +1,1062 @@
...
+filled in with the result. As an example, consider the signature of the Async
+equivalent of `In_channel.read_all`. <idx>Deferred.t</idx>
+
+<link rel="import" href="code/async/main.mlt" part="3" />
+
+ We first load the Async package in the toplevel using `#require`, and then
+open the module. Async, Like Core, is designed to be an extension to your
+basic programming environment, and is intended to be opened.
+
+ A deferred is essentially a handle to a value that may be computed in the
```
yminsky: Is the leading space in front of each graf important? Or is it just an artifact of the conversion?

book/02-variables-and-functions.md
```diff
@@ -0,0 +1,886 @@
...
+<link rel="import" href="code/variables-and-functions/main.mlt" part=
+"28" />
+
+The syntactic role of an operator is typically determined by its first
+character or two, though there are a few exceptions.
+[Table2_1](02-variables-and-functions.html#table2_1){data-type=xref} breaks
+the different operators and other syntactic forms into groups from highest to
+lowest precedence, explaining how each behaves syntactically. We write 
+`!`... to indicate the class of operators beginning with `!`.
+
```
yminsky: Nothing to do, but it's worth noting that these tables are pretty hard to read. Fixable with the auto-align functionality in emacs, I think.

samoht: There are various ways to express tables in pandoc markdown: https://pandoc.org/MANUAL.html#tables

Happy to switch to something more readable (even if it is a bit of more work). We can also convert them manually directly.

book/18-concurrent-programming.md
yminsky: The purpose of these is to generate an index. So they shouldn't typically be rendered in the final text directly, but instead are fed to an index-generation process. We did this as part of generating the original O'Reilly book (actually, they wrote the index for us), and it would be nice not to lose that work, and indeed, to be able to help in updating and generating the index as we work on the book.

[event loops]{.idx} seems like an OK syntax for these.

book/02-variables-and-functions.md
yminsky: Agreed. The syntax is actually not  a problem from my perspective; it's just that it would be nice to have alignment. Agreed we can just add this by hand later.

samoht:
I think this is ready to merge. The only drawback is that this needs a fairly recent of pandoc (at least 2.*) which are not packaged yet on ubuntu/debian (but is in brew). The easiest way to install it is to follow what the Dockerfile is doing:

```
curl -OL https://github.com/jgm/pandoc/releases/download/2.1.3/pandoc-2.1.3-1-amd64.deb && sudo dpkg -i pandoc-2.1.3-1-amd64.deb
```

yminsky:
Yeah, fedora also had an older Pandoc. Anyway, I'm still happy to merge it, and get the updated version installed somehow.

PR #2939, opened 2018-04-03
accepted test changes, fixing some old bugs in output

yminsky:
Looked at the output as well, which looks cleaner.


examples/code/async/search_with_configurable_server/run_search_with_configurable_server.errsh
```diff
@@ -1,9 +1,8 @@
   $ jbuilder build search_with_configurable_server.exe
-%% --non-deterministic
+% --non-deterministic
```
samoht: it's a bit strange that there is only one `%%` symbol here, seems that a regression in the pretty-printer. I'll have a look.

examples/code/async/search_with_configurable_server/run_search_with_configurable_server.errsh
```diff
@@ -1,9 +1,8 @@
...
   $ ./_build/default/search_with_configurable_server.exe -servers localhost,api.duckduckgo.com "Concurrent Programming" OCaml
   (monitor.ml.Error (Unix.Unix_error "Connection refused" connect 127.0.0.1:80)
    ("Raised by primitive operation at file \"src/unix_syscalls.ml\", line 937, characters 17-76"
     "Called from file \"src/deferred1.ml\", line 20, characters 40-45"
     "Called from file \"src/job_queue.ml\", line 159, characters 6-47"
     "Caught by monitor Tcp.close_sock_on_error"))
-@@ exit 1
```
samoht: and it is strange that that `@@ exit 1` disappeared as well ...

PR #2941, opened 2018-04-03
Fix bug in variants chapter

yminsky:
The error must have come from the initial move to Base, which hid polymorphic compare. I just used Int. and Float. explicitly.


yminsky:
This is based on the accept-test-fixes branch, so it shouldn't be reviewed until that one is merged, I suppose.

yminsky:
I'll merge without review, since it's simple, and Anil is on vacation.

PR #2945, opened 2018-04-10
get rid of chapter numbers

yminsky:
Chapter order is now entirely determined by the listing in the toc file.


samoht:
The `jbuild.inc` files seem not to be vert stable between various regeneration (I suspect it depends on the ordering of `readdir`). I will fix that later on;

So LGTM.

PR #2956, opened 2018-05-08
Use a unique toplevel jbuild file for testing code examples

samoht:
This helps with test running `jbuilder` themselves (like some tests in the `code/json` directory) and will help to write the testing chapter that @yminsky is working on.

Note: the downside of segregating tests this way is that `jbuild build @examples/code/.../runtest` doesn't work anymore. It's possible to add back individual test commands if needed.


samoht:
It seems that the test chapter now works (almost) fine with these changes. No more crazy recursive outputs, but that `cram` binary is not found in path. Will dig into this later.

samoht:
I have added individual test targets. To run `examples/code/async/echo/run_echo.sh` use:

```
jbuilder build @runtest-code-async-echo-run_echo.sh
```

I still need to investigate why `cram` is not in the path of the tests.

PR #2957, opened 2018-05-10
cram: add a missing flush

samoht:


samoht:
I am merging this quickly as it fixes an annoying issue in the cram tool (that was hit by the `testing` branch).

PR #2959, opened 2018-05-22
removing reference to non existing log_entry

3kwa:
log_entry is referenced in the text but does not appear in the guided tour.

Elected not to mention that `let ... in` is used above in the second definition of `ratio` as it uses the more *complicated* `open Float.O` rather than a simpler new binding definition such as the one described in local_let.mlt.


yminsky:
This has now been fixed in master.

PR #2961, opened 2018-05-31
Updated to Ocaml v4.02  and Core v0.11

christophe-riolo:
I updated the examples to ocaml 4.02 and core 0.11.

If I review correctly only the Ncurses example in the Foreign Interfaces chapter does not work, I know too little about the foreign interfaces to make it work.

For the async_graphics example in the chapter on classes, we need an updated async_graphics library. so I took https://github.com/drewsdunne/async_graphics and added an opam file, so we can pin to it : https://github.com/christophe-riolo/async_graphics

I also updated the scripts, and the Makefile accordingly (because of the required "chapter" option)


christophe-riolo:
I realized I was working from the wrong branch ...

PR #2967, opened 2018-06-01
Remove data-type=sect2 and data-type=sect3 attributes in HTML headers

samoht:
These doesn't seem to be useful at all, so probably better to just remove them as they could easily go out of sync.


samoht:
> These doesn't seem to be useful at all

The CI seems to say otherwise :-) I will investigate..

avsm:
needs a rebase, sorry :)

samoht:
Rebased again

PR #2968, opened 2018-06-01
css: fix display of warning boxes

### LINKED ISSUE #2880: Rendering bug with warning sections
Warning sections are mis-rendered in the current book. Here's one from the maps chapter in v1:

image

And here's the same one in v2:

image

There's no bounding box or mark of the warning, and the text just flows abruptly into the next graf.

samoht:
Fix #2880

The new rendering:

screen shot 2018-06-01 at 10 42 35

That commit also change the font size of "note" boxes to be the same as the "sidebar" and "warning" ones. I still have no idea what the sidebar boxes are about, and why they are treated differently (e.g. they are in a `<aside>` tag instead of a `<div>`) :-) We could probably improve all of that by treating all these boxes almost the same. Just let me know what you prefer.

### DIFF
--- a/static/css/app.css
+++ b/static/css/app.css
@@ -1262,26 +1262,46 @@
      padding-top: 0;
      border-top: 0;
 }
+
+ /* Notes & Warnings */
  section[data-type=chapter] div[data-type=note] {
      background: #fdedd4;
      border: 1px solid #f39c12;
+ }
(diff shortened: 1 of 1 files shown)


avsm:
Let's go with this for now -- I can't remember the difference.

PR #2969, opened 2018-06-01
Correction of some exercises with a failed build

christophe-riolo:
Title is explicit and most changes are also. Couple of notes : 

- async_graphics was incompatible with core 0.11 but @drewsdunne did all the job of updating it. I added an updated opam file and it was good to go, so I pinned my repository using opam 2 `pin-depend`. It requires opam 2 though.

- I changed the names of the variables because `x` and `y` were already bound and the error message was not representative of what a user would see in: 
https://github.com/realworldocaml/book/compare/master...christophe-riolo:master#diff-675913d9cf8eb716b98f5f1e08c0a3b3

### DIFF
--- a/examples/code/command-line-parsing/md5_let_syntax2/md5.sh
+++ b/examples/code/command-line-parsing/md5_let_syntax2/md5.sh
@@ -1,3 +1,4 @@
   $ dune build md5.exe
+%% --non-deterministic
   $ ./_build/default/md5.exe 5 ./_build/default/md5.exe
   8ff8c
--- a/examples/code/command-line-parsing/md5_let_syntax2/md5.sh
+++ b/examples/code/command-line-parsing/md5_let_syntax2/md5.sh
@@ -1,3 +1,3 @@
   $ dune build md5.exe
   $ ./_build/default/md5.exe 5 ./_build/default/md5.exe
-  45a17
+  8ff8c
--- a/examples/code/files-modules-and-programs/freq-with-missing-def/build.errsh
+++ b/examples/code/files-modules-and-programs/freq-with-missing-def/build.errsh
@@ -1,6 +1,6 @@
   $ dune build freq.bc
         ocamlc .freq.eobjs/counter.{cmo,cmt} (exit 2)
-  (cd _build/default && /home/christophe/.opam/rwo/bin/ocamlc.opt -w -40 -g -bin-annot -I .freq.eobjs -I /home/christophe/.opam/rwo/lib/base -I /home/christophe/.opam/rwo/lib/base/caml -I /home/christophe/.opam/rwo/lib/base/shadow_stdlib -I /home/christophe/.opam/rwo/lib/sexplib0 -I /home/christophe/.opam/rwo/lib/stdio -no-alias-deps -o .freq.eobjs/counter.cmo -c -impl counter.ml)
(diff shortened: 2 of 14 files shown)


samoht:
I think that's a great idea to switch the CI to use opam2, I am just waiting for @avsm to push the final image for the new [ocaml containers infrastructure](https://github.com/ocaml/infrastructure/wiki/Containers). 

Once we have that in place, I plan to turn on `make test` on one run of the CI, so we can tests that all dependencies can install correctly and that the examples compile fine (which means fixing all the examples too).

christophe-riolo:
I must say that `make test` never ended for me, at some point it stopped doing anything, for hours, but straight after cloning. I tested manually with `jbuilder build @runtest-...`.
Maybe it will run properly for you 🙂

samoht:
Which version of ocaml are you using? The `-short-path` options used to be very slow when used with Core/Async (and it still is until #2970 is done). But `make test` usually ends for me, it just print a lot of errors ...

christophe-riolo:
I found the problem, it was with netcat, that froze waiting for an end of file, so it would depend on the version of echo/netcat I guess. With this option it should work ... At least it does for me :)

Also I found another exercises that slipped through my attention !

christophe-riolo:
Edit: I see that there are two versions of netcat (at least: GNU and openbsd) and they don't have compatible options. Openbsd needs `-N` and GNU need `-c` and they do the same thing. And looks like there might be other versions out there that don't need it. Figures.

christophe-riolo:
I managed to get the ncurses examples to run, so I'm adding them :)

samoht:
Would you mind rebasing your PR on master? I've just merged #2998 which fixes uses dune and fixes a few more tests.

samoht:
I am using 4.06.1

christophe-riolo:
Ok, I tested again, fixed one more test.
The netcat issue bothers me a bit as previously said because it depends on which version of netcat you have, but there it is.

examples/code/command-line-parsing/md5_let_syntax2/md5.sh
```diff
@@ -1,3 +1,3 @@
   $ dune build md5.exe
   $ ./_build/default/md5.exe 5 ./_build/default/md5.exe
-  45a17
+  8ff8c
```
samoht: the command should be prefixed by `%% --non-deterministic`

examples/code/files-modules-and-programs/freq-with-missing-def/build.errsh
```diff
@@ -1,6 +1,6 @@
   $ dune build freq.bc
         ocamlc .freq.eobjs/counter.{cmo,cmt} (exit 2)
-  (cd _build/default && /home/christophe/.opam/rwo/bin/ocamlc.opt -w -40 -g -bin-annot -I .freq.eobjs -I /home/christophe/.opam/rwo/lib/base -I /home/christophe/.opam/rwo/lib/base/caml -I /home/christophe/.opam/rwo/lib/base/shadow_stdlib -I /home/christophe/.opam/rwo/lib/sexplib0 -I /home/christophe/.opam/rwo/lib/stdio -no-alias-deps -o .freq.eobjs/counter.cmo -c -impl counter.ml)
+  (cd _build/default && /home/christophe/.opam/rwo/bin/ocamlc.opt -w @a-4-29-40-41-42-44-45-48-58-59-60-40 -strict-sequence -strict-formats -short-paths -keep-locs -g -bin-annot -I .freq.eobjs -I /home/christophe/.opam/rwo/lib/base -I /home/christophe/.opam/rwo/lib/base/caml -I /home/christophe/.opam/rwo/lib/base/shadow_stdlib -I /home/christophe/.opam/rwo/lib/sexplib0 -I /home/christophe/.opam/rwo/lib/stdio -intf-suffix .ml -no-alias-deps -opaque -o .freq.eobjs/counter.cmo -c -impl counter.ml)
```
samoht: can you replace that line with `...` to avoid machine-dependant strings in the test outputs?

christophe-riolo:
Changes done, I didn't know we could do that thank you :)

XVilka:
Looks like this PR is now irrelevant and can be closed, right?

christophe-riolo:
Most likely, I'll close it

Issue #2809, opened 2017-12-23
Layout hard to read


skaller:
The online layout is almost unreadable. The fonts are too large and the line spacing way too large. The contents column is not movable and scales proportionally to the width of the browser defeating any attempt to get enough text on the screen to actually read. The right hand margin is way too large. 

Code examples which require scrolling are no good either. [Argg .. and opan can't install utop because lwt is broken]

NathanReb:
I'm closing as I think the style has changed since then. Please feel free to re-open if you feel like RWO still is unreadable!

PR #2974, opened 2018-06-01
Compact style

samoht:
I have trying to reduce a bit the empty spaces in the current style and to reduce the visual noise of having so many different font sizes. It's not perfect, but I feel it's a bit better compared to what it was before. There are still two many different fonts, size, and noise to my taste, but the current CSS is complex enough and my time is limited ...

Before:

screen shot 2018-06-01 at 18 48 17

After:

screen shot 2018-06-01 at 18 48 23

On the variant chapter (where there are subsections):

Before:

screen shot 2018-06-01 at 18 50 59

After:

screen shot 2018-06-01 at 18 51 03

### DIFF
--- a/book/install.html
+++ b/book/install.html
@@ -38,7 +38,7 @@
 
   <p>Let's get started.</p>
 
-  <h1>Getting OPAM</h1>
+  <h2>Getting OPAM</h2>
 
   <p>The easiest way to install OPAM is through your OS's package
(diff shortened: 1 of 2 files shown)


samoht:
This is related to #2809

avsm:
Needs a rebase but lgtm

samoht:
Rebased

PR #2983, opened 2018-06-18
[WIP] have one file per chapter

samoht:
This is a work-in-progress branch to convert all the chapters to only use one file for both code and examples: instead of have one `md` file and multiple scripts per chapter, now everything is one one `mlt` file where text and command is interleaved (using a literate programming style).

This PR implements only one part of the workflow: convert exiting files to `mlt`. The remaining work is to:
- update ocamltopexect to skip text (e.g. `md` attributes)
- update the tests to execute the mlt files instead of having individual chapter tests
- update the HTML generator to read the mlt file

Note: one way to be sure we don't break anything would be to keep the current text and examples files, but generate them from the chapter files. Once we are happy with the result, we can remove the intermediate step (or not).


samoht:
Also, something needs to be done for the current mlt "headers", e.g. code which is run before any part, like [here](https://github.com/realworldocaml/book/blob/master/examples/code/fcm/main.mlt#L1-L7)

yminsky:
Maybe we can handle that with something like #hide true / #hide false around those segments?

y

samoht:
Closing in favor in #2987

PR #2985, opened 2018-07-19
[Variables and Functions] replace List.dedup with List.dedup_and_sort

savil:
This is in the section on left-associativity of `|>`. The code snippet above refers to `List.dedup_and_sort` but the text calls it as `List.dedup`.


yminsky:
Thanks!

PR #2987, opened 2018-07-19
Use one md file per chapter

samoht:
Manual testing with [mdx](https://github.com/realworldocaml/mdx) seems to work fine (at least on the class chapter).

TODO:
- [ ] update the auto-gen code to add `runtest` command using `mdx test`
- [ ] update the HTML generator to convert code block to lisible HTML fragments


samoht:
Note: the shape of build rules to generate in order to test the md files is:

```
(alias
 ((name    runtest)
  (deps    (classes.md))
  (action  (progn
            (run mdx test --prelude=prelude.ml classes.md)
            (diff? classes.md classes.md.corrected)))))
```

book/guided-tour.md
```diff
@@ -154,13 +203,24 @@ slightly awkward `/.` operator. In the following example, we open the
 are designed to be used in this kind of context. Note that this causes the
 standard int-only arithmetic operators to be shadowed locally.
 
-<link rel="import" href="code/guided-tour/main.mlt" part="6.1" />
+```ocaml
+# ;;
```
yminsky: This looks messed up --- the ordering of the lines is backward

yminsky:
From a quick glance this looks asotnishingly good.

Is the idea that some of these snippets are sourced from particular external files, and those files are inlined into the file via the generation of a corrected file? So, the actual toplevel stuff we can do by editing the .md file, but the modification of a shell snippet must be done by editing the original shell file. Does that sound right?

book/guided-tour.md
samoht: Thanks, fixed

samoht:
> Is the idea that some of these snippets are sourced from particular external files, and those files are inlined into the file via the generation of a corrected file? So, the actual toplevel stuff we can do by editing the .md file, but the modification of a shell snippet must be done by editing the original shell file. Does that sound right?

Not everything is centralised in the `md` file: the `ocaml` *and* the `sh` code blocks are interpreted differently by `mdx test`. The ocaml fragments are run via a custom ocaml toplevel, and the the shell scripts are run in the directory specified with `dir=...` when defining the block:

    ```sh dir=<path>
    $ ...
     ```

It could be passed as an argument of `mdx test` using the `--root` argument, but it's a bit less convenient to use here.

book/command-line-parsing.md
```diff
@@ -44,7 +44,17 @@ a file, applies the MD5 one-way cryptographic hash function to the data, and
 outputs an ASCII hex representation of the result: [MD5 one-way cryptographic
 hash function]{.idx}[command-line parsing/basic approach to]{.idx}
 
-<link rel="import" href="code/command-line-parsing/md5/md5.ml" />
+```ocaml
```
yminsky: I'm very confused by this example: The source here is an ml, not an mlt, but it's not obvious to me how I would know that. I would assume that I would need some metadata here to keep track of the location of the file that this example comes from. 

What happens in the workflow if I edit the text here? Does it get reverted next time I build, or does it flow the other way, causing the shell examples later to change to reflect running this new code?

book/command-line-parsing.md
```diff
@@ -55,12 +65,16 @@ for optional flags and positional arguments, including documentation, the
 types they should map to, and whether to take special actions such as pausing
 for interactive input if certain inputs are encountered.
 
-### Defining an anonymous argument {#anonymous-arguments}
+### Defining an anonymous argument anonymous-arguments
```
samoht: This is wrong

book/classes.md
```diff
@@ -175,13 +420,29 @@ use explicit modules for illustration, but the process is similar when we
...
+  let make init = new stack init
+end
+```
 
 We have multiple choices in defining the module type, depending on how much
 of the implementation we want to expose. At one extreme, a maximally abstract
 signature would completely hide the class definitions:
 
-<link rel="import" href="code/classes/class_types_stack.ml" part="1" />
+```ocaml file=examples/code/classes/class_types_stack.ml,part=1
```
samoht: Here an example of the new promotion feature of `mdx`: `mdx test` will *not* try to run this code block anymore, it will just check that the code block is in sync with the `ml` file (and it will generate a `.corrected` file if that's not the case).

samoht:
The tooling integration with `mdx` is now complete (with the dev version). I have moved all chapter files into their own subdirectories, to ease integration with dune, so one could do `cd book/variants && dune runtest`.

Most of the chapter will need minor manual tweaking to make the tests pass. Here a summary of what I've looked so far:

- [ ] classes
- [ ] command-line-parsing
- [ ] compiler-backend
- [ ] compiler-frontend
- [ ] concurrent-programming
- [ ] data-serialization
- [ ] error-handling
- [ ] files-modules-and-programs
- [ ] first-class-modules
- [ ] foreign-function-interface
- [ ] functors
- [ ] garbage-collector
- [ ] guided-tour
- [ ] imperative-programming
- [ ] json
- [ ] lists-and-patterns
- [ ] maps-and-hashtables
- [ ] objects
- [ ] parsing-with-ocamllex-and-menhir
- [ ] prologue
- [ ] records
- [ ] runtime-memory-layout
- [ ] variables-and-functions
- [ ] variants

samoht:
I've converted a few more chapters; hope to finish the conversion before the end of the week.

avsm:
My bad; I did a couple of minor PR merges that conflict with this. Wont push to trunk until the merge!

book/garbage-collector/README.md
```diff
@@ -442,17 +459,65 @@ Let's see this for ourselves with a simple test program. You'll need to
...
+  Bench.make_command tests |> Command.run
+```
 
 This program defines a type `t1` that is mutable and `t2` that is immutable.
 The benchmark loop iterates over both fields and increments a counter.
 Compile and execute this with some extra options to show the amount of
 garbage collection occurring:
 
-<link rel="import" href="code/gc/barrier_bench/dune" />
+```scheme
```
avsm: Not a blocker for this PR, but given that this is an OCaml tool, it might be nice to explicitly call this `dune` rather than `scheme`.  That would let us do dune-specific parsing in the future such as link to the online docs for particular stanzas.

samoht:
I believe this is now ready to merge 🎉 

```
$ make test
dune runtest
```

And the HTML seems to look similar than it was before. I believe regressions can be fixed in a subsequent PR.

PR #2989, opened 2018-08-24
Fixing polymorphic `>`example in variants.md

brooklynrob:
In paragraph describing use of the `>` in polymorphic types, the example I think should cite ``[> `Float of float | `Int of int]`` not ``[> `Int of string | `Float of float]``.


yminsky:
Thanks for the fix! Merged.

PR #2993, opened 2018-09-18
Make native executable extension consistent

btmills:
:wave: Thanks for working on this in the open! I'm just starting to learn, and it's been a great resource.

Line 943 calls the native executable `sum.exe`, while 947 calls it `sum.native`, so I updated the latter to make them consistent.


yminsky:
Thanks for the fix!

PR #2998, opened 2018-10-01
Use Dune

samoht:
Also fix `make test`; on my machine this fixes all the tests and I've tried to add ellipsis (`...`) where dirnames and filenames appear in the test output (mainly on compilation errors).


samoht:
Note: after this, it would be possible to turn on `make test` on PRs to check regression in code examples.

samoht:
I am merging this one as @yminsky approved #2987 which includes the same commits.

PR #2999, opened 2018-10-02
Fix typo in Hash Tables, "Base" instead of "Core"

joelonsql:
Discussion: https://discuss.ocaml.org/t/possible-typo-base-or-core/2650


yminsky:
Thanks!

PR #3002, opened 2018-10-02
Fix md5.ml - use expected version of do_hash

joelonsql:
When reaching the "Argument Types" section in the book,
the declared version of "do_hash" takes two arguments (hash_length and filename),
so the "do_hash file" doesn't work and gives error:
Error: This expression has type string but an expression was expected of type
         int

Fix by basing the code example on the latest definition of "command".


XVilka:
I think this one is not relevant anymore - all examples already look good in this chapter.

PR #3004, opened 2018-10-05
Fix int_interval.ml, missing module ... end

joelonsql:
Without this, the code example in test_interval.ml fails with "Error: Unbound module Int_interval".


yminsky:
Fixed here: https://github.com/realworldocaml/book/commit/b30dea0f49ea46e08de6a1e170a1b4130da166da

PR #3005, opened 2018-10-05
Fix test_interval.mli, add missing module sig

joelonsql:


yminsky:
As described elsewhere, not the correct fix.

PR #3011, opened 2018-10-08
Fix run_search_with_timeout_no_leak.sh

joelonsql:
This is the corresponding text in the book:

"Now, if we run this with a suitably small timeout,
we'll see that one query succeeds and the other fails reporting a timeout:"

So one of the queries should timeout.


yminsky:
Interesting. Did you rerun the test and have it timeout accordingly? Or did you just modify the text of the output?

yminsky:
Ping.

yminsky:
Fixed in master.

PR #3016, opened 2018-10-09
Reduce horizontal coverage for clickable area in links

TyOverby:
Previously, the clickable portion of the ToC and Navbar links was the entire horizontal width of their container instead of just the part of the link that has visible text.  While browsing the book, I mis-clicked a few times on what I thought was empty space, but was actually the continuation of a link.

`display: table-cell` is a great replacement for `display: block` because it keeps all the vertical layout the same, while also having a traditional "span style" horizontal fill.

### DIFF
--- a/static/css/app.css
+++ b/static/css/app.css
@@ -1000,7 +1000,7 @@ h6 {
      text-transform: uppercase;
 }
  .toc-full a {
-     display: block;
+     display: table-cell;
 }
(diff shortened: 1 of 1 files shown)


samoht:
I am far for being an expert in CSS expert, so thank you very much for your patch :-) Feel free to fix other issues what you would find!

PR #3032, opened 2018-11-17
Small test fixes to cause "make test" to run cleaner

yminsky:
Accepted some test changes that make sense, marked some sections as non-deterministic. Note that there are some tests that print out times that are non-deterministic only because the rendering depends on the time-zone of the host machine, even though the time itself is deterministic. We should fix this at some point by making tests pick a default time-zone to run in. But for now, this seems good enough.

### DIFF
--- a/README.md
+++ b/README.md
@@ -41,7 +41,7 @@ There are three main repositories for the book:
   uses the scripts and examples repositories to compile the HTML
   site online.
 
-All of the code and examples are built using OCaml 4.06.1.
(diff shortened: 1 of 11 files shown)


samoht:
Which version of the compiler are you using? The `Core_kernel` prefixes appear for me on 4.07.1.

yminsky:
I'm now using 4.07.1, and have reverted all those changes. I also updated rwo.opam to add async_graphics, which was necessary to make things build cleanly.

PR #3041, opened 2018-12-19
Fix invalid urls

colin4124:


book/prologue/README.md
```diff
@@ -249,7 +249,7 @@ the language and about libraries tools and techniques that will help you be a
 more effective OCaml programmer. But it's no replacement for API
 documentation or the OCaml manual and man pages. You can find documentation
 for all of the libraries and tools referenced in the book
-[online](https://realworldocaml.org/doc).
+[online](https://ocaml.janestreet.com/ocaml-core/).
```
yminsky: @avsm: what do you think about this? Pointing at the JS docs seems wrong, since there are other libraries we use in RWO. Maybe that's OK as a temporary move, but we should really get docs generated for the RWO libraries specifically...

avsm: I'll take a look at generating the docs ourselves. I'm very close to having a full "rwo duniverse" working as the last difficult library to port is en route now with https://github.com/ocamllabs/ocaml-ctypes/pull/588 

Once that's done, the docs are as simple as a `dune build @doc` from a cronjob. 

I'll merge this particular PR now as the current URL is wrong, and update it when those docs are live.  Thanks for this PR @colin4124!

colin4124: You are welcome

PR #3043, opened 2018-12-26
"String.uppercase" --> "String.uppercase_ascii"

brooklynrob:
String.uppercase has been deprecated. Replacing "String.uppercase" with "String.uppercase_ascii" in Deferred.bind example early in README.md of "Concurrent Programming with Async" chapter that introduces >>= use, as well as two other examples further down.

### DIFF
--- a/book/concurrent-programming/README.md
+++ b/book/concurrent-programming/README.md
@@ -134,7 +134,7 @@ uppercase version of its contents.
 # let uppercase_file filename =
     Deferred.bind (Reader.file_contents filename)
       (fun text ->
-         Writer.save filename ~contents:(String.uppercase text))
+         Writer.save filename ~contents:(String.uppercase_ascii text))
 val uppercase_file : string -> unit Deferred.t = <fun>
 # uppercase_file "test.txt"
 - : unit = ()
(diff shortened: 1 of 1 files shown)


brooklynrob:
After this pull request was approved by @yminsky I realized that String.uppercase has been deprecated in the OCaml standard library, but not in [Base](https://github.com/janestreet/base/). 

I wanted to point it out as I believe RWO assumes all exercises are done using Base, not [Ocaml stdlib](https://github.com/ocaml/ocaml/blob/trunk/stdlib/string.mli). In that case the original version using `String.uppercase` not `String.uppercase_ascii` before my PR would be correct.

For context, see the definition in the [OCaml stdlib](https://github.com/ocaml/ocaml/blob/trunk/stdlib/string.mli):
```
val uppercase : string -> string
  [@@ocaml.deprecated "Use String.uppercase_ascii instead."]
(** Return a copy of the argument, with all lowercase letters
   translated to uppercase, including accented letters of the ISO
   Latin-1 (8859-1) character set.
   @deprecated Functions operating on Latin-1 character set are deprecated. *)
```
and (vs).

```
val uppercase_ascii : string -> string
(** Return a copy of the argument, with all lowercase letters
   translated to uppercase, using the US-ASCII character set.
   @since 4.03.0 *)
```

Here's what the corresponding functions look like in the [Base standard library](https://github.com/janestreet/base/blob/master/src/string.mli):

```
(** Operates on the whole string using the US-ASCII character set,
    e.g. [uppercase "foo" = "FOO"]. *)
val uppercase : t -> t
val lowercase : t -> t

```

PR #3053, opened 2019-01-21
Using orignal polymorphic compare operators

colin4124:
Using orignal polymorphic compare operators in Caml instead of integer compare operators in Base.

Function `find`'s interface is defined as polymorphic in `.mli`.
```ocaml
val find   : ('a, 'b) t -> 'a -> 'b option
```

But `find`'s implement in `.ml` is inferred to `val find : (int, 'a) t -> int -> 'a option`
```ocaml
open Base

let find t key =
  List.find_map t.buckets.(hash_bucket key)
    ~f:(fun (key',data) -> if key' = key then Some data else None)
```

Because the type of the equality operator that’s open in Base is `int -> int -> bool` . So I use the original polymorphic operator in such way:

```ocaml
open Base

let find t key =
  List.find_map t.buckets.(hash_bucket key)
    ~f:(fun (key',data) -> if Caml.(key' = key) then Some data else None)
```

### DIFF
--- a/examples/code/imperative-programming/dictionary.ml
+++ b/examples/code/imperative-programming/dictionary.ml
@@ -21,10 +21,9 @@ let create () =
 
 let length t = t.length
 
-let find t key =
+let find t key ~equal =
   List.find_map t.buckets.(hash_bucket key)
-    ~f:(fun (key',data) -> if Caml.(key' = key) then Some data else None)
-
+    ~f:(fun (key', data) -> if (equal key' key) then Some data else None)
 
 [@@@part "3"];;
 let iter t ~f =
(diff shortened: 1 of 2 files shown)


yminsky:
That's a good find, but I don't really want to fix it this way. I'd rather modify the mli so that it takes an equality function as an argument, similar to how the List module works in Base.

https://github.com/janestreet/base/blob/master/src/list.mli#L432

Would you be up for adjusting the PR to do that?

colin4124:
@yminsky Thanks for your suggestion.  Taking an equality function as an argument rather than using a polymorphic compare operator is more  elegant. 

I have adjusted,  does it do that?

XVilka:
Ping? Looks good to merge.

samoht:
This needs to be reflected in the corresponding README too

PR #3057, opened 2019-02-03
Update link to examples

naartjie:
Update the examples link to point to https://github.com/realworldocaml/book/tree/master/examples.

I'm not sure if there is a long terms permalink strategy, and also if the paragraph text will need to be updated, the current version of the code snippets don't have any clickable header links.


yminsky:
Fixed by https://github.com/realworldocaml/book/commit/1b189babe3754fba5470ef0feb126bd3c0cecf25

yminsky:
(Thanks for pointing out the issue!)

PR #3059, opened 2019-02-16
guided-tour: Remove mention of missing log_entry var

softmoth:


yminsky:
Thanks for pointing this out.  I'm going to fix this problem slightly differently, by #3129.

PR #3073, opened 2019-05-06
Couple of tweaks

czystyl:
I don't want to open issue per thing so that's why I'm taking all of the things here. 

If any of all fixes are not needed just let me know :) I'm still reading the book and that's the reason why it's a draft.

Fixes:
- [x] double `and`
- [x] ```File "time_converter.ml", line 6, characters 30-35:
Error (warning 3): deprecated: Core.stdin
[since 2016-04] Use [In_channel.stdin]```


czystyl:
We can merge it to not deal with big PR - if more fixes will come. 

Feel free to merge it.

PR #3089, opened 2019-07-05
Fix syntax for multiple type constraints on a signature.

mdebski:
Following the example used to lead to a unhelpful syntax errors from the
compiler.


yminsky:
Nice catch. Thanks.

PR #3099, opened 2019-07-22
Fix type error of the example in chapter 4

sangwoo-joh:
According to the explanation, it should be `Username` rather than `Hostname`.


yminsky:
Thanks for the fix!

PR #3100, opened 2019-08-01
Duniverse: Build and test with duniverse

NathanReb:
This is a work in progress.

So far we have a working `duniverse` and we started fixing the broken tests so that we can re-enable testing in the CI before merging this.

As the rule of thumb, we leave the failures that are due to `mdx` or `duniverse` issues because we need to address those on our side, we fix the easy failures and we skip the tests that seem broken beyond repair and require chapter authors intervention.


NathanReb:
@yminsky this sets up a duniverse and re-enable the tests in CI. I'll open subsequent PRs to disable/fix the tests that are currently broken.

Note that because the base branch is lagging a bit behind master some of the commits here aren't actually part of this PR.

PR #3103, opened 2019-08-04
Add missing pattern for id

k4rtik:


yminsky:
@avsm This one has been sitting around for a while. Should we close it?

k4rtik:
@yminsky do you think the CI needs to be run again after merging recent changes from the main branch?

I think only the maintainers can initiate that.

avsm:
Thanks for the report @k4rtik!  the id token was deliberately left out of the example parser since we don't parse unquoted strings. Adding it would require adding a parser token and trigger an unused variable warning. I've added a note to the chapter that unused regexps are fine, in 37414a61cb32437cc29e8b886a5b4bd4a6a25dd9.

PR #3105, opened 2019-08-07
Duniverse: Fix Files, Modules and Programs chapter

NathanReb:
Depends on #3100, I'm leaving this as draft until it's merged! Note that only the last two commits are relevant to this PR.

This fixes the tests in the "Files, Modules and Programs". Most of the changes were straightforward as it was just out of date error message. There's a few things worth noting though:
- There are trailing whitespaces from the new `mdx` here as well
- `mdx` now adds `;;` after the `[@@@part ...]` attributes, I'm not sure if there's a rationale behind this but I reported it here: https://github.com/realworldocaml/mdx/issues/147
- I disabled a shell snippet with an `ocamlfind` invocation. The current dune port for findlib/ocamlfind has a few issues and I don't think it builds the `ocamlfind` executable or at least doesn't do so in a `dune` friendly way. Furthermore, it is not properly installed and lacks an initial `findlib.conf`. That needs to be fixed in the dune port although a fix upstream would probably be better.

### DIFF
--- a/book/files-modules-and-programs/README.md
+++ b/book/files-modules-and-programs/README.md
@@ -117,7 +117,7 @@ need a somewhat more complex invocation to get them linked in: [OCaml
 toolchain/ocamlc]{.idx}[OCaml toolchain/ocamlfind]{.idx}[Base standard
 library/finding with ocamlfind]{.idx}
 
-```sh dir=../../examples/code/files-modules-and-programs/freq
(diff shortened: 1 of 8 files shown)


avsm:
For the record: the dune port of ocamlfind is split into two parts: findlib (the library) and ocamlfind (the binary)

PR #3106, opened 2019-08-07
Duniverse: Fix Records chapter tests

NathanReb:
Depends on #3100, I'm leaving this as draft until it's merged! Note that only the last 4 commits are relevant to this PR.

Not much going on here:
- The toplevel was missing a `#require "re"`
- I removed the deprecated `#require "re.posix"`
- I moved from the deprecated `Re.get` to `Re.Group.get`

### DIFF
--- a/book/records/README.md
+++ b/book/records/README.md
@@ -46,7 +46,7 @@ just think of them as a simple pattern language you can use for parsing a
 string.
 
 ```ocaml env=main
-# #require "re.posix"
+# #require "re,re.posix"
 # let service_info_of_string line =
     let matches =
       Re.exec (Re.Posix.compile_pat "([a-zA-Z]+)[ \t]+([0-9]+)/([a-zA-Z]+)") line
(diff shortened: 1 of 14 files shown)


NathanReb:
Hmm this doesn't work anymore for some reason, I'll have a look!

NathanReb:
We found the issue and have a working fix, we just need to get some dune port released to opam-overlays and it should be good!

`findlib` was silently failing because there was no `seq` library available. I'm not sure why this wasn't properly picked up when setting up the duniverse.

NathanReb:
Fixed and updated PR description

yminsky:
I don't know how to review this.  Github thinks 6k files have changed, which I presume is not correct...

Julow:
That's because this PR is based on top of https://github.com/realworldocaml/book/pull/3100. It Fixes trivial things like generated files and add the duniverse, which is the source code of each dependencies.

yminsky:
Is there a nice way in GitHub for me to read these diffs properly?

Julow:
Yes, when #3100 is merged into the `duniverse` branch we will rebase this PRs on top of it and the diffs will be cleaner.

PR #3109, opened 2019-08-08
Duniverse: Fix json chapter

Julow:
On https://github.com/realworldocaml/book/pull/3100

I changed `Yojson.Basic.json` to `Yojson.Basic.t` but this is not required.
Otherwise it's syncing ml files and specifying a dependency to atdgen.


Julow:
This chapter is not fixed yet.

`atdgen` is not building and `menhir` is required but not compatible with dune.

NathanReb:
It's ready for review, the problematic tests are disabled.

I haven't look at it but for the `atdgen` issue we probably need to take a deeper look at how to support that in `mdx`, it might be a bug in the new `require_package` feature.
This might also require some tricks to add executables to the path but I think `dune` already takes care of that.

avsm:
Looks fine to merge; can look at the `skip` tests later on

PR #3114, opened 2019-08-08
Duniverse: Fix parsing with ocamllex and menhir chapter

Julow:
On #3100 

`Command.Spec.file` does not exists. I replaced it with `string`, is that correct ?


avsm:
`file` appears to have disappeared indeed -- `string` is fine!

PR #3119, opened 2019-08-14
Fix concurrent programming

NathanReb:
This required a few changes, here's a summary:
- The code depended on `cohttp.async` which has now moved to the separate `cohttp-async` package. This also resulted in a small change in a function signature as the API evolved since then.
- Old and new versions of snippets from the chapter co-existed in `search_with_error_handling.ml`, I removed the outdated parts.
- Because `examples/code` is marked as ignored and folders there have their own dune-project, running a dune command inside that directory won't find the libraries from the duniverse, unless we add dependencies over the packages providing them. This is not ideal and I think a proper fix would be to rethink how we organize the repo around the examples. Marking them as vendored and removing the local `dune-project` could be an option as it should then be able to find libraries from the duniverse while still preventing dune from resolving aliases in that folder by default.


avsm:
lgtm

PR #3120, opened 2019-08-14
Duniverse: Fix foreign function interface

NathanReb:
This is still a work in progress atm and a few issues remain.

Before simply skipping the test I'd like to take the time to figure out why exactly it's not working. So far here are the various errors I still get:
1. For some reason some bindings don't work in the toplevel with the following:
```
# let time = foreign "time" (ptr time_t @-> returning time_t)
-val time : time_t Ctypes_static.ptr -> time_t = <fun>
+File "_none_", line 1:
+Error: Reference to undefined global `Foreign'
```
2. `corebuild` invocation fail because of the missing ocamlfind binary, I guess this can be fixed in the dune port of ocamlfind. I'm wondering whether we need corebuild though.
```
``sh dir=../../examples/code/ffi/ncurses
 $ corebuild -pkg ctypes-foreign ncurses.inferred.mli
+Failure: ocamlfind not found on path, but -no-ocamlfind not used.
+[2]
```
3. When building example no implementation is found for `ctypes-foreign`:
```
$ dune build qsort.exe
+Error: No implementation found for virtual library "ctypes-foreign" in
+/home/nathan/code/rwo-book/_build/install/default/lib/ctypes-foreign.
+-> required by executable qsort in dune:2
+[1]
```


NathanReb:
Note that I fixed a few issues, especially `ctypes.foreign` vs `ctypes-foreign` and a few out of sync examples/toplevel snippets.

avsm:
Thanks! I'll take a look at the skipped issues in a separate PR

PR #3133, opened 2019-08-24
remove unnecessary require declarations

yminsky:


yminsky:
@avsm Do you understand why these require statements are necessary? Removing them causes CI to break (but oddly, things didn't break on my own machine), but I don't understand what they're for.  After all, there are some packages that are used (like Async!) that don't have to be listed with an explicit require.

avsm:
It looks like they add an explicit `(package)` dependency to the dune file, for when it can't be inferred elsewhere.  I'll take a closer look as to why the CI is diverging from your laptop, which shouldn't be the case.

NathanReb:
The `require-package` labels are required because that's how we let `mdx` know that the execution of this code depends on the presence of external executables provided by those packages.
`dune` should then make them available in the `PATH`.

@Julow knows more about this feature.

In this particular case, I believe that to compile `search.ml` you need to have `yojson` and `textwrap` available. Because no other snippet in this `.md` file depends on them we have to specify them explicitly. The reproducibility issue might come from the fact that some other rules depending on `yojson` and `textwrap` were executed before, resulting in them being present in the `_build/install/default/lib`.

In short, removing them result in a race condition.

NathanReb:
I also think that is tied to how we setup the example directories which prevents dune from finding the dependencies in the duniverse and thus requiring us to make them available in the `_build/install` folder somehow.

I'll try to investigate that further.

PR #3140, opened 2019-09-01
Update command

yminsky:
Update the command directory, using dune exec and making a bunch of things deterministic.  This involved fixing some type-errors that had crept in.


yminsky:
@avsm The above seems to fail because we don't have cryptokit in the duniverse. (though it oddly works on my laptop.)  Can we get cryptokit dune-ified?

avsm:
I'm updating this PR to use `Core.Md5` instead of cryptokit.

yminsky:
That seems fine too!

avsm:
I don't have permissions to push to your fork, so I'll close this and have opened https://github.com/realworldocaml/book/pull/3145 as a followup

PR #3145, opened 2019-09-02
Update CLI chapter to remove cryptokit and use dune exec

avsm:
followup to #3140 

@yminsky, is the let-syntax example used anywhere yet in the text?

### DIFF
--- a/book-pkgs
+++ b/book-pkgs
@@ -20,7 +20,6 @@ core_bench
 core_extended
 core_kernel
 cppo
-cryptokit
 domain-name
 easy-format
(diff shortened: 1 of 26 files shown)


avsm:
I'm mainly wondering if it's wise to encourage the use of `string` for the file contents. But I didn't want to pull in all the Bigstring complexity.  Perhaps revisiting this in the Async chapter to demonstrate Async file i/o for a very large input might illustrate the point

yminsky:
I'm not sure what your questions about let syntax is.  It's definitely used in the chapter.

Issue #3052, opened 2019-01-20
Single-directory chapters


yminsky:
It would be really great if we could move directories to be single chapter affairs. i.e., instead of having examples be in a separate example directory, moving them so they are in subdirectories of the directory with the README for the chapter in question.

Maybe this is already possible now, and it's just a matter of moving things around and changing paths. Is that true?

Issue #3139, opened 2019-09-01
Promotion runs in the wrong direction for non-toplevel code (and infinite promotion loops)


yminsky:
Right now, promotion goes from the README file to the examples, which seems like the wrong direction.  Generally, one wants to go and fix ordinary OCaml code in the ordinary source file, since that's where you get proper feedback from the tools.

Also, there's some weird behavior where I sometimes see an infinite promotion loop, specifically when using parts and having spaces between the parts markers --- it can end up wanting to insert newlines endlessly.  Maybe changing the direction of promotion will fix this bug as well.

avsm:
Agreed; I don't ever want to be editing OCaml code in a Markdown context.

avsm:
Confirmed I'd like this changed as well. When editing #3140, I:

- changed the OCaml files in `examples/code/command-line-parsing/md5/md5.ml`
- ran `dune build --root examples/code/command-line-parsing/md5 md5.exe` to check they work
- ran `dune runtest book/command-line-parsing` -- and promotion happened in the wrong direction, and my changes in `examples/code/command-line-parsing/md5/md5.ml` were proposed to be overwritten by the version in the book.

NathanReb:
Totally agree we're doing it the wrong way here.

I think that'll make even more sense once we moved the examples in the corresponding chapter's folder and make sure `runtest` actually build all examples that don't contain intentional errors.

I'll open a PR with that workflow for one chapter and we can discuss it further there!

yminsky:
Can we make those changes independently? Certainly, I don't want to wait
for the disabled tests to be turned back on and fixed. This workflow really
makes getting those fixes in harder!

I don't nice waiting for the directory reorganization, as long as we do
that soon.

NathanReb:
Yes of course! Also note that I plan to do that reorganization real soon anyway!

PR #3156, opened 2019-09-04
Single-directory chapter

Julow:
Fixes https://github.com/realworldocaml/book/issues/3052

This PR moves examples from `examples/code` into the corresponding chapter directory (`book/chapter/examples`).

If a chapter has more than one directory in `examples/code/$example_dir` they are moved to `book/chapter/examples/$example_dir`.


book/compiler-frontend/README.md
```diff
@@ -1090,7 +1090,7 @@ This module defined bindings for the Ncurses library. First, compile the
 interfaces with `-bin-annot` so that we can obtain the `cmt` and `cmti`
 files, and then run `ocp-index` in completion mode:
 
-```sh dir=../../examples/code,source-tree=../../examples/code/ffi,skip
+```sh dir=../foreign-function-interface/examples,source-tree=../foreign-function-interface/examples/ffi,skip
```
Julow: `compiler-frontend` uses an example from `foreign-function-interface`.

book/maps-and-hashtables/README.md
```diff
@@ -49,7 +49,7 @@ Let's consider an example of how one might use a map in practice. In
 we showed a module `Counter` for keeping frequency counts on a set of
 strings. Here's the interface:
 
-```ocaml file=../../examples/code/files-modules-and-programs/freq-fast/counter.mli
+```ocaml file=../files-modules-and-programs/examples/freq-fast/counter.mli
```
Julow: Same for `maps-and-hashtables` and `files-modules-and-programs`.

book/compiler-frontend/README.md
NathanReb: My take on this is that we should duplicate them now thus allowing them to diverge in the future. I think depending on files that aren't within the current directory doesn't cope well with `dune` and it makes maintining the chapters more complex than it should.

Any thought on that @yminsky @avsm ?

NathanReb:
Have you tried removing the gadt examples? If they aren't used anywhere we should simply get rid of them, we can still browse the history if we ever want to bring them back.

NathanReb:
Splitting this into one PR per chapter + a wrap up PR would probably ease the review but if it's too much hassle we can do with a single PR!

NathanReb:
Nevermind the commit view is good enough!

Julow:
The diff is simple, it's just renames and path changes and a single PR will make easy to do other small things.

Removed `gadts`.

The `examples` directory is empty so I removed it, what should I do with the `UNLICENSE` file ?
Should I copy it into every `book/chapter/examples` ? If there were code already in some `book/chapter/examples` should it be an exception ?

book/compiler-frontend/README.md
avsm: Duplicating is just fine in this case.

Julow: I wonder if duplicating will cause confusion. The text before this block says "Let's refer back to our Ncurses binding example from  [...]" and the block starts with `(cd ffi/ncurses [...]` (this path may have to be updated). Readers should be able to find the source code easily.

NathanReb: I think it'd be okay to duplicate. If we want the code to be strictly identical we can enforce it with a dune `diff` rule for the duplicates in `book/dune` for instance.

No solution is perfect, even the current one as if we rewrite a chapter A without removing the example it's still possible that a later chapter B references out of date examples from A and it wouldn't be enforced by the CI anyway.

NathanReb:
Could you fix the duplicate thing so we can merge and I can start working on #3139 ?

Julow:
I copied the problematic examples.
Is there some text to rewrite to explain the new scheme to readers ?

NathanReb:
Good point! I think the README needs an update but that's true regardless of this PR as it mentions external repos we don't use anymore. Let's open a separate issue for that and merge this as it is!

Julow:
This PR removes the `examples/UNLICENSE` file that were applied to every examples I moved.
Should I copy it into every example directories, changing the license of files that already in `book/*/examples`, or to individual directory/file touched by this PR ?
Otherwise, would it be possible to move this file to the root of the project and add the list of file it apply to ?

PR #3169, opened 2019-09-15
Update install instructions

avsm:


book/install.html
```diff
@@ -238,15 +229,16 @@ <h2>Configuring OPAM</h2>
 system  C system                      System compiler (4.00.1)
 --     -- 4.04.2                      Official 4.04.2 release
 --     -- 4.05.0                      Official 4.05.0 release
---     -- 4.06.0                      Official 4.06.0 release
+--     -- 4.07.1                      Official 4.07.1 release
+--     -- 4.07.1                      Official 4.07.1 release
```
yminsky: Why did 4.07 show up twice?

avsm: good catch, pushed a fix in de39a00 along with fixes to some other references to ancient ocaml versions

Issue #3079, opened 2019-05-22
It looks like your example formatter "eats" the `<fun>` that is part of utop's/toplevel's output


zhenya1007:
It looks like your example formatter "eats" the `<fun>` that is part of utop's/toplevel's output.  For example, the type of `ratio` function, as printed by utop/toplevel ends up being shown as: 

```
val ratio : int -> int -> float = 
```

I am quite certain it is _actually_ printed as 

```
val ratio : int -> int -> float = <fun>
```

I am guessing your example formatter "thinks" that `<fun>` is an unknown HTML/XML tag, and "eats" it.

avsm:
It looks like this might be a highlight.js issue, since the output html has `&lt;fun&gt;` which is then turned into `<fun></fun/>` after the highlighting pass.

PR #3176, opened 2019-09-22
actually fix html toplevel syntax highlighting

avsm:
my fix to c74ef8bf896c5ea21c97ccfb339ce23278420213 accidentally
put the wrong version of the JavaScript in.  This one removes
some debug sprinkles and introduces the patch that fixes
the #3079 reported issue of `<fun>` not rendering from toplevel
output.


avsm:
The live site works now. I'm slightly confused about the sequence of events, since I'm sure I checked the live site when I merged in the fix for #3079. But anyway, it's all in now.

PR #3193, opened 2019-10-11
imperative-programming fixed few typos

vitaliel:


book/imperative-programming/README.md
```diff
@@ -367,7 +367,7 @@ You can see these in action:
 
 ```ocaml env=ref
 # let x = ref 1
-val x : int Caml.ref = {Base.Ref.contents = 1}
```
yminsky: This is not a typo! That's actually what the toplevel spits back.  You can't just edit this, since it breaks the CI.

vitaliel: ok, in my  utop version 2.3.0 (using OCaml version 4.07.1)! it shows without Caml.

vitaliel: Reverted the line

PR #3196, opened 2019-10-11
Add chapter title to html.head.title

vitaliel:
Also removed duplicated head tag, before it was `<head><head>`

### DIFF
--- a/bin/lib/book.ml
+++ b/bin/lib/book.ml
@@ -6,9 +6,9 @@ let (/) = Filename.concat
 (******************************************************************************)
 (* HTML fragments                                                             *)
 (******************************************************************************)
-let head_item ?(page_title=None) () : Html.item =
+let head_item chapter_title : Html.item =
   let site_title = "Real World OCaml" in
-  let t = match page_title with
+  let page_title = match chapter_title with
       | None -> site_title
       | Some t' -> sprintf "%s - %s" t' site_title
   in
@@ -19,7 +19,7 @@ let head_item ?(page_title=None) () : Html.item =
       "name","viewport";
       "content","width=device-width, initial-scale=1.0"
     ] [];
-    title [`Data t];
+    title [`Data page_title];
     link ~a:["rel","stylesheet"; "href","css/app.css"] [];
     link ~a:["rel","stylesheet"; "href","css/prism.css"] [];
     script ~a:["src","js/min/modernizr-min.js"] [];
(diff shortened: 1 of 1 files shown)


bin/lib/book.ml
```diff
@@ -6,15 +6,20 @@ let (/) = Filename.concat
 (******************************************************************************)
 (* HTML fragments                                                             *)
 (******************************************************************************)
-let head_item : Html.item =
+let head_item ?(page_title=None) () : Html.item =
+  let site_title = "Real World OCaml" in
+  let t = match page_title with
```
yminsky: "t" is a bit of a confusing name here.  How about title?

yminsky: Also, head_item is called only once, so there's no point in making the extra argument optional.  Better to have a mandatory labeled argument, e.g., "head_time ~page_title".  Then you can pass page_title in as an explicit argument.

yminsky: page_title isn't a great name for the value here, since the title you end up computing is, when you think about it, the title of the page.  I think chapter_title might be a better name.

bin/lib/book.ml
```diff
@@ -138,10 +143,11 @@ let next_chapter_footer next_chapter : Html.item option =
     front page and only chapter pages contain links to a next chapter,
     so these are additional arguments. *)
 let main_template ?(next_chapter_footer=None)
+    ?(page_title=None)
```
yminsky: Change the definition of the optional arguemnt to be:

```
?page_title
```

rather than explicitly providing None as the default value. That way, when you call main_template, you can pass it an ordinary string (rather than a string option), which will simplify the call point.

bin/lib/book.ml
```diff
@@ -6,15 +6,20 @@ let (/) = Filename.concat
 (******************************************************************************)
 (* HTML fragments                                                             *)
 (******************************************************************************)
-let head_item : Html.item =
+let head_item chapter_title : Html.item =
```
yminsky: I think it's better to label this argument.

bin/lib/book.ml
```diff
@@ -138,10 +143,11 @@ let next_chapter_footer next_chapter : Html.item option =
     front page and only chapter pages contain links to a next chapter,
     so these are additional arguments. *)
 let main_template ?(next_chapter_footer=None)
+    ?page_title
```
yminsky: I'm pretty sure this should be chapter_title, not page_title.  page_title will be computed from this.

PR #3200, opened 2019-10-18
Objects: typo in subtyping example (**) instead of (**.)

arnau:
The subtyping example uses `**` instead of `**.`. This PR fixes it.

### DIFF
--- a/book/objects/README.md
+++ b/book/objects/README.md
@@ -370,7 +370,7 @@ end
 type circle = < area : float; radius : int >
 
 let circle r = object
-  method area = 3.14 *. (Float.of_int r) ** 2.0
+  method area = 3.14 *. (Float.of_int r) **. 2.0
   method radius = r
 end
 ```
--- a/book/objects/examples/subtyping.ml
+++ b/book/objects/examples/subtyping.ml
@@ -13,7 +13,7 @@ end
 type circle = < area : float; radius : int >
 
 let circle r = object
-  method area = 3.14 *. (Float.of_int r) ** 2.0
+  method area = 3.14 *. (Float.of_int r) **. 2.0
   method radius = r
 end
 


yminsky:
The change looks good, but the markdown wasn't updated.  You need to run "dune runtest; dune promote" to fix that.

arnau:
@yminsky I'm afraid I don't understand why CI is failing. I ran `make test` locally and went well. Status 0, no observable issues. 🤔

book/files-modules-and-programs/README.md
```diff
@@ -153,33 +153,19 @@ itself. [OCaml toolchain/dune]{.idx}
 
 ```sh dir=examples/freq-dune
 $ grep -o '[[:alpha:]]*' freq.ml | ./_build/default/freq.bc
```
yminsky: This looks wrong, and is also what the test is complaining about.  I'm not sure how this change got in here, but it looks like it should be reverted.

arnau: ugh, I have no idea. Let me fix this

arnau:
@yminsky I removed the odd change. This piece of code depends on `grep` and mine, under macos outputs something different to what the CI and I assume an up to date Linux does.

If it helps in any way:

* macos Mojave 10.14.6
* grep (BSD grep) 2.5.1-FreeBSD
* GNU bash, version 3.2.57(1)-release (x86_64-apple-darwin18)

arnau:
Ok, I installed grep (GNU grep) 3.3 and now the output is as expected by the tests.

yminsky:
Looking at this more, it looks like there's a bigger problem, which is that somewhere along the way as we redid the tooling, those examples got set up in a way that they don't compile.  I went and did a larger cleanup of the chapter, which includes your change, and a bunch of other small tweaks, here:

https://github.com/realworldocaml/book/pull/3201

Since that PR subsumes this one, I'm going to close this.  Thanks for pointing out the issue!

PR #3203, opened 2019-10-20
Added warning contents to Records chapter

antonlogvinenko:
Code examples that generate warnings usually include the warning text, so it seems like warning 40 here should be too. The warning is also discussed right after the code sample.


yminsky:
I don't think this is a sensible change, which is why CR broke. You're editing the output of a command, which you can't actually do.  You'd have to change the setup so the warning was emitted when the code in question is actually executed.

yminsky:
Looking at this more, it seems like the warning just doesn't show up anymore by default, due to upstream compiler changes.  So, I'm updating the text in https://github.com/realworldocaml/book/pull/3206

antonlogvinenko:
Oops, sorry, didn't know that!

PR #3206, opened 2019-10-26
fix small issues with records chapter

yminsky:


antonlogvinenko:
Oops, sorry, didn't know that!

PR #3218, opened 2019-11-14
Import output command from mdx

### LINKED ISSUE #3213: Import HTML and LateX generation from mdx
Currently we use `mdx output` command to generate the book HTML. What that command does is:
1. It strips the mdx specific labels from the markdown
2. Add custom styling to OCaml code blocks
3. Invoke pandoc with a specific set of options to generate our HTML

All this is very tightly tied to our needs and it would make more sense to import this code in the repo so we can iterate over it faster and remove it from `mdx` as it is unrelated to `mdx`'s role.

Since mdx exposes a public library we can start by importing the `mdx output` command code as is.

clecat:
This pull request is the first step of this [issue](https://github.com/realworldocaml/book/issues/3213):
It imports the code of the `output` command of mdx.

The code is almost the same, only a few parts were cleaned.

Fixes #3213


bin/bin/dep.ml
```diff
@@ -146,8 +146,8 @@ let process_md ~toc book_dir =
         let html = chapter ^ ".html" in
         sprintf {|(rule
   (targets %s)
-  (deps    %s)
-  (action  (run mdx output %%{deps} -o %%{targets})))|}
+  (deps    (:x %s) ../bin/bin/output.exe)
```
samoht: it's a bit strange to depend on the private name here. Does it work if you depend on `%{exe:rwo-output}` instead? (or whatever the dune syntax is for this)

clecat: You are right, it works perfectly and is way better !
However, the other parts of `dep.ml` are using the private name as well, do you want me to change that too in this PR ?

samoht: you can add a commit to fix the other parts of `dep.ml` if you feel like it, but I don't think that could block the merge of that PR.

clecat: You are right, it works perfectly and is way better ! (the right syntax is `bin`, `exe` seems to look in the current folder only)
However, the other parts of dep.ml are using the private name as well, do you want me to change that too in this PR ?

bin/bin/dep.ml
```diff
@@ -146,8 +146,8 @@ let process_md ~toc book_dir =
         let html = chapter ^ ".html" in
         sprintf {|(rule
   (targets %s)
-  (deps    %s)
-  (action  (run mdx output %%{deps} -o %%{targets})))|}
+  (deps    (:x %s) %%{bin:rwo-output})
```
NathanReb: I think a more explicit name than `x` would probably be better!

Also, with dune, you don't need to explicit declare the dependency on `rwo-output` since it knows how to build the first argument to `run`, it understands this dependency on its own.

bin/bin/dune
```diff
@@ -1,7 +1,7 @@
 (executables
  (package      rwo)
- (names        app       dep)
- (public_names rwo-build rwo-dep)
+ (names        app       dep       output)
```
NathanReb: I think we should give up on a trying to align names and public names, especially since they aren't here!

As a matter of fact we should even let dune format our dune files.

NathanReb: We should also try to come up with a better name than `output` as it's a bit vague, especially now that it's a standalone executable and not a subcommand of `mdx`.

What do you think about `convert_md`/`rwo-convert-md`?

NathanReb:
Also can you please add a "Fixes #3218" in the PR description so it closes the issue automatically once it get merged?

bin/bin/dep.ml
clecat: Okay, thank you for this information, I'll  change that

bin/bin/dune
clecat: Ho, so that's why those spaces were there, I could not figure it, I'll get rid of that.

For the name, rwo-convert_md looks good to me

avsm:
Good to go!

PR #3219, opened 2019-11-16
removed some skips

yminsky:


yminsky:
@NathanReb This one is breaking because atdgen isn't installed.  Can you maybe separately add atdgen into the duniverse, so we can remove these skips?

yminsky:
Ping @NathanReb.  Can we get the duniverse upgraded so that this PR will build?

NathanReb:
Sure, I'll take a look today!

NathanReb:
I gave it a quick look and atdgen is already in the duniverse. I think the issue here is that we can't invoke the binary from a shell block and hope it to pick up the binary from the duniverse without having dune set some env variables before.

I.e. that would work with `dune exec -- atdgen ...` but we obviously don't want that in the book as it doesn't reflect how readers would use `atdgen`.

I'll try to see what we can do here!

yminsky:
@NathanReb Any progress on this issue?

yminsky:
@NathanReb I've caught this up with master.  If we migrate this chapter to the new example workflow, will that fix the present problem? Or is there something else we need to do?

avsm:
dup of #3367

PR #3222, opened 2019-11-18
dockerfile: speedup and use latest dune 1.11.4

avsm:
remove several layers of workdirs that weren't needed, and upgrade
to dune 1.11.4

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -10,17 +10,13 @@ RUN git -C /home/opam/opam-repository pull origin master && opam update -uy
 # install non-OCaml dependencies
 COPY Makefile /home/opam/src/.
 RUN make depext
-RUN opam install dune=1.11.3
+RUN opam install dune=1.11.4
 
 #install pandoc
-WORKDIR /tmp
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -10,17 +10,13 @@ RUN git -C /home/opam/opam-repository pull origin master && opam update -uy
 # install non-OCaml dependencies
 COPY Makefile /home/opam/src/.
 RUN make depext
-RUN opam install dune=1.11.3
+RUN opam install dune=1.11.4
 
 #install pandoc
-WORKDIR /tmp
```
NathanReb: Removing this workdir leads to a permission error, we should probably explicitly ask curl to write to `/tmp` with `-o /tmp/pandoc-2.1.3-1-amd64.deb`

XVilka:
By the way, the latest pandoc is https://github.com/jgm/pandoc/releases/tag/2.8.0.1
2.1.3.1 is way too old.

avsm:
No longer needed, closing

PR #3232, opened 2019-12-04
support dune 2.0.0

avsm:
part of the journey towards #3227 and 4.08 support


avsm:
For the record, this was mostly running `dune upgrade` over the duniverse.  When https://github.com/ocaml/opam-repository/pull/15444 is merged we should be in good shape to update the duniverse

avsm:
With many thanks to @NathanReb for debugging the issue that was blocking this (a missing `seq` library in duniverse), this is now ready for merge.

PR #3233, opened 2019-12-07
update to 4.08

### LINKED ISSUE #3227: update to 4.08.1

avsm:
this diff seems suspiciously short.  Need to investigate why more tests aren't affected by OCAML_ERROR_STYLE changing.
*Update:* this is because dune 2.0 terminates the build after the first promotion failure, instead of trying all possible rules. Investigating why, but not a blocker for this PR.

Closes #3227


avsm:
This should now pass the main Travis CI; @yminsky could you glance over and merge when happy? 

The other ocaml-ci should also now pass once https://github.com/ocurrent/ocaml-ci/pull/81 is reviewed and merged, so ignore that failure for now.

avsm:
I filed https://github.com/realworldocaml/book/pull/3233 to track the promotion bug

PR #3234, opened 2019-12-09
Add and document a way to upgrade the duniverse

### LINKED ISSUE #3216: Document the procedure to upgrade the duniverse
The procedure to upgrade the duniverse should be documented, likely in the README or in a new HACKING so that anyone can do it when it needs to be done.

NathanReb:
Fixes #3216.

This is a short term solution but it works. We have to maintain the list of packages in the `Makefile` by hand but that would be true of a solution based on the opam or dune-project file as well for now.

We now have to new `make` targets:
- `duniverse-init` which will re-resolve our deps and therefore update the dune-get to the latest available versions
- `duniverse-upgrade` that depends on the previous target and calls `duniverse pull` to also upgrade the content of the duniverse folder.

I grep-ed the dune files in `book/` and `bin/` to figure out our list of direct dependencies so I managed to get rid of a few unnecessary repos in our duniverse.

I also updated the opam file so the required ocaml and dune versions in there reflect what we use in development.

A longer term fix would involve changes to `duniverse` (the tool). Ideally we'd like the `dune-project` package stanza to act as the source of truth regarding direct dependencies. We can imagine that dune could maintain the list of our dependencies automatically for us using the information from rule, library, and executable stanzas.

Let me know what you think!


avsm:
diff is too big but could you also delete `book-pkgs` if thats not already done.

NathanReb:
I just pushed a fix so the CI should go green now!

PR #3238, opened 2019-12-11
test if ocaml-ci works

avsm:
This just reverts a constraint failure in duniverse; needs to be
fed upstream before it can be merged


avsm:
closed for https://github.com/realworldocaml/book/pull/3241

PR #3243, opened 2019-12-19
Upgrade to OCaml 4.09 and Core v0.13.0

avsm:


yminsky:
Do you know why ocaml-ci failed?

avsm:
Looks like ocaml-ci failed due to exceeding the maximum number of layers in the container image.  This is in turn due to the vast number of pins required for depext calculation in the duniverse, whereas the Travis CI currently hardcodes this depext information.  I've opened up https://github.com/ocurrent/ocaml-ci/issues/85 to address this upstream, so it's not a blocker here.

PR #3253, opened 2020-01-09
Update book/records/README.md

imbsky:


avsm:
I don't think this is right. The text refers to `field punning` and `label punning` as two features, not a plural binding to the `both` earlier in the sentence.

imbsky:
Oh, I see!

PR #3254, opened 2020-01-13
Handwrite mdx rules

NathanReb:
This removes the generation of dune rules through `ocaml-mdx rule` as some of its mechanisms were opaque.

We'll hopefully be able to replace them with the first iteration of dune's mdx stanza soon. That should help us get rid of some of the unneeded verbosity.

You'll note that I dropped the `runtest-all` alias which wasn't very useful in practice since it was never run on the whole book but rather on some blocks, the first time they were written or once in a while to keep them up to date. As a replacement, one can simply momentarily remove the `non-deterministic` label from a block to achieve the same effect.

I will allow setting the `--non-deterministic` option of mdx through an env variable so that it can easily be run on a single chapter with `MDX_NON_DETERMINISTIC=true dune runtest book/<chapter>` without us having to maintain similar aliases.

I also removed the `--root` option passed to the mdx invocation in the JSON chapter and fixed the path in the chapter accordingly so that it follows the same conventions as the other chapters and can be extended with new examples more easily if need be.


yminsky:
Let's talk about this at today's roundup.  I don't really understand why this is a good idea, since it replaced some hard-to-understand rules for generating dune files with some hard-to-understand dune files!

NathanReb:
To sum offline discussions: we'll wait until we have the `mdx` stanza in dune instead as those rules would be hard to maintain!

PR #3255, opened 2020-01-14
Upgrade duniverse

NathanReb:
I wanted to upgrade mdx but decided to upgrade the whole duniverse instead.

Nothing major here except for Lwt which is bumped to 5.0.0 but is seems to work just fine!


book/data-serialization/README.md
```diff
@@ -596,11 +596,7 @@ Uncaught exception:
   (Of_sexp_error "int_of_sexp: (Failure int_of_string)"
    (invalid_sexp not-an-integer))
 
-Raised at file "duniverse/sexplib0.v0.13.0/sexp_conv.ml", line 203, characters 30-72
-Called from file "read_foo.ml", line 5, characters 2-3
-Called from file "read_foo.ml", line 3, characters 0-71
-Called from file "read_foo.ml", line 11, characters 4-60
-Called from file "duniverse/base.v0.13.0/src/exn.ml", line 102, characters 6-10
+...
```
NathanReb: You'll note that I added this so that the output is not affected by the duniverse's structure.
If this part of the error messag eis actually relevant and needs to be there, we'll need to make improvements
on mdx to fix it!

PR #3256, opened 2020-01-14
Remove some unneeded folders in examples

NathanReb:
Depends on #3254 

The json and command-line-parsing had all their examples grouped in a single folder within their example folder.

This PRs gets rid of this extra folder.

PR #3257, opened 2020-01-14
Synchronize more files from examples

### LINKED ISSUE #3151: Sync dune files fragment from examples to markdown files
There are quite a few fragments of dune files in the book and they aren't validated atm.

It would be nice to at least keep the ones extracted from examples in sync with the actual dune file, the same way we wish to keep ocaml fragments in sync with the code in the examples!

Eventually being able to validate dune files fragments in general would be great but I'm not even sure we have such snippets that aren't extracted from an actual dune file in the examples folder.

NathanReb:
Fixes #3151 

Depends on #3256 

This uses the new arbitrary file sync from mdx to sync some non-ml files from the examples in some chapters (dune, .json, etc..)


NathanReb:
I was able to sync quite a few files, even `.ml` files that previously weren't!

I just skipped the [dune file](https://github.com/realworldocaml/book/blob/master/book/first-class-modules/examples/query_handler_loader/dune) from the first-class-modules chapter as it wasn't clear whether the chapter or the book needed fixing!

NathanReb:
Just rebased this, should be ready for review!

NathanReb:
Hmm looks like github didn't receive the message that the build succeeded, I restarted the travis build hoping it will fix it.

PR #3267, opened 2020-02-07
Use z instead of x in Nesting Lets with Let and In

shakthimaan:
In the "Nesting lets with let and in" section, the value of "x" is '3' instead of the following:
```
  utop> x;;
  Line 1, characters 1-2:
  Error: Unbound value x
```
If one continuously evaluates the chapter examples in a utop session, the value of 'x' gets set as '3' from the "Tuples" section:
```
  let (x,y) = a_tuple;;
  val x : int = 3
  val y : string = "three"
```
Hence, an unused variable "z" is used to illustrate the let binding.

### DIFF
--- a/book/guided-tour/README.md
+++ b/book/guided-tour/README.md
@@ -804,18 +804,18 @@ within which the new variable can be used. Thus, we could write:[let
 syntax/nested let binding]{.idx}
 
 ```ocaml env=local_let
-# let x = 7 in
-  x + x
+# let z = 7 in
+  z + z
 - : int = 14
 ```
 
 Note that the scope of the `let` binding is terminated by the
-double-semicolon, so the value of `x` is no longer available:
+double-semicolon, so the value of `z` is no longer available:
 
 ```ocaml env=local_let
-# x
+# z
 Line 1, characters 1-2:
-Error: Unbound value x
+Error: Unbound value z
 ```
 
 We can also have multiple `let` statements in a row, each one adding a new
--- a/book/guided-tour/README.md
+++ b/book/guided-tour/README.md
@@ -803,7 +803,7 @@ including a function body. The `in` marks the beginning of the scope
 within which the new variable can be used. Thus, we could write:[let
 syntax/nested let binding]{.idx}
 
-```ocaml env=local_let
+```ocaml env=main
 # let z = 7 in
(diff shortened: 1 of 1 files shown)


yminsky:
This seems like a reasonable change, but if the goal is to make it all work in the same toplevel environment (which seems nice enough), it would be good to change the environment from local_let to main, and make sure that everything keeps working.

shakthimaan:
I was reading the HTML page, and was trying out the examples, one-by-one in utop.  Hence, the value of `x` was 3 as initialized. I am not sure if this is how newbies are expected to try the examples, or you prefer them to execute the code examples that are provided with the book. Also, only after looking at the book source did I notice that `ocaml env=local_let` is used. If this is the intended use, then this could be made explicit when viewing the the HTML? For various sections in a chapter, it will be good to run them stand-alone as much as required, and not depend on any prior state, especially when learning.

yminsky:
I think the expectation is that people will mostly do things in order.  I think this particular one is a little digression (and, now that I look at it, it looks like it's malformatted.  It's supposed to be a highlighted box that's clearly not part of the ordinary flow).  Anyway, if you want to make it all work together, that seems reasonable, but we might as well zip up the environments to check that it really works.  Can you make that change and rerun the tests and make sure it's all OK?

shakthimaan:
CMIIW, we would like to use `ocaml env=main` for the block with the variable `z`, and run the tests to make sure everything is fine?

yminsky:
Correct!

shakthimaan:
The output of `make test`:
~~~~{.sh}
$ make test
dune runtest
  ocamlmklib duniverse/ctypes.0.16.0+dune/src/ctypes-foreign-base/dllctypes_foreign_base_stubs.so,duniverse/ctypes.0.16.0+dune/src/ctypes-foreign-base/libctypes_foreign_base_stubs.a
Unknown option -Wl,--no-as-needed
      ocamlc duniverse/core_kernel.v0.13.0/src/md5_stubs.o
md5_stubs.c:11:32: warning: ‘-pedantic’ is not an option that controls warnings [-Wpragmas]
   11 | #pragma GCC diagnostic ignored "-pedantic"
      |                                ^~~~~~~~~~~
~~~~

PR #3270, opened 2020-02-07
Provide example for string option * string in context

shakthimaan:
It will be good to give an example for "string option * string" in the "Options" section for completion. Moreover, it might take some time for the reader to reach Chapter 7: Error Handling.
```
  utop> Some "File";;
  - : string option = Some "File"

  utop> (Some "File", "exists");;
  - : string option * string = (Some "File", "exists")
```


yminsky:
I don't think the extra example really improves the explanation here.  The guided tour is meant to have a brisk pace, to introduce you to things without fully explaining all of them. The extra example, I think, weighs the prose down.  The issue will be covered in more detail in the error handling chapter anyway, as is indicated.

PR #3273, opened 2020-02-11
Use MDX stanza

### LINKED ISSUE #3271: Simplify mdx rules
We're currenlty working on adding an mdx stanza to dune which will greatly simplify the way we run it for out test suite.

Once it's ready we should upgrade the dune language version and use it instead of the rule generation scheme we currently rely on.

NathanReb:
Fixes #3271 

This PR replaces the hand written mdx rules with uses of the new mdx stanza.

it required a dune lang bump so we're using dune 2.4 now.

Among some of the noticeable changes, dune rules target are now read-only, which means I had to tweak the pdf generation rules a bit since some of them were overwriting files they depended upon.


book/command-line-parsing/dune
```diff
@@ -1,15 +1,9 @@
...
- (name   runtest)
- (deps   README.md)
- (action (diff dune.inc dune.gen)))
-
-(include dune.inc)
+(mdx
+ (files README.md)
+ (packages
+  core
+  mdx
```
avsm: How come the `mdx` stanza depends on `mdx`? :-)

NathanReb: That's because of the two binary issue, I'll ask the dune team if there's a better way to work around that until we have a native toplevel!

PR #3278, opened 2020-02-24
Update MDX

NathanReb:
This depends on #3274 so I'm opening it as a draft PR until it's merged.

It updates our vendored MDX to the latest version.

The points worth noting are:
- It will be compatible with dune's mdx stanza when `dune.2.4.0` is released
- It allows using comments for part delimiters in OCaml files we want to include using MDX
- The `--direction` option has been removed since we weren't using it anymore
- MDX does not unset `INSIDE_DUNE` by default, which means we have to set it by hand if we really want that behaviour. Only a single code block is affected by this change.


NathanReb:
I am a bit puzzled by the CI failures, I can't reproduce them locally.

NathanReb:
Great, I'll try to figure out why those tests are failing in the CI and not locally!

NathanReb:
It turns out that bumping the dune version fixed the issues we had!

PR #3279, opened 2020-03-05
Included (modes byte) to first dune file

tottenhamjm:
You now have to explicitly tell dune that it can compile to bytecode with (modes byte) in the dune file. Updated chapter 4 to reflect this in order to avoid:
`Error: Don't know how to build freq.bc` from `dune build freq.bc`.


yminsky:
We need to actually upgrade the version of dune we're invoking in the dune config files for this to work. That's a bigger job that's being worked on separately.

yminsky:
This PR should solve the issue once it's released 

https://github.com/realworldocaml/book/pull/3321

Moving to use native code compiler by default, rather than adding the bytecode clause.

PR #3284, opened 2020-03-25
[Lists and Patterns] Use meaningful variables in List.fold

shakthimaan:
In the `Using the List Module Effectively` section in the `Lists and Patterns` chapter, instead of saying:
```
  utop> List.fold ~init:[] ~f:(fun list x -> x :: list) [1;2;3;4];;
  - : int list = [4; 3; 2; 1]
```
Can it be more specific to put across the use of fold as shown below?
```
  utop> List.fold ~init:[] ~f:(fun accum head -> head :: accum) [1; 2; 3; 4];;
  - : int list = [4; 3; 2; 1]
```
Output of `dune runtest`:
```
$ dune runtest
$
```


yminsky:
Thanks! Closing in favor of #3322, which uses variable names more like the rest of the chapter.

PR #3285, opened 2020-03-26
[Files Modules and Programs] Update freq-dune example to build with dune 2.0

shakthimaan:
In the "Where is main?" section, the "dune build freq.bc" does not work with dune 2.0. The present file looks like:
```
   (executable
     (name      freq)
     (libraries base stdio))
```
The updated dune file should be:
```
   (executable
     (name freq)
     (modes byte exe)
     (libraries base stdio))
```
Output of `dune runtest` is clean.
Reference: https://github.com/ocaml/dune/issues/3017#issuecomment-572502753


yminsky:
Resolving this differently in #3321

PR #3287, opened 2020-03-27
update duniverse, including ocaml-print-intf

avsm:


avsm:
cc @NathanReb

avsm:
This is a consistent duniverse now, but fails to build due to the mdx interface changes. Could you take a look at that @NathanReb ?

NathanReb:
@avsm if you don't mind I'll hold the mdx update back for a bit so we can get the `ocaml-print-intf` part merged faster!

NathanReb:
There seem to be both good and bad in the mdx erros but I'd like to properly investigate this!

NathanReb:
I pushed a commit rolling back mdx to 1.6.0!

avsm:
Thanks, I'll do a separate commit to update ocaml-print-intf

PR #3289, opened 2020-04-01
Replace corebuild with ocaml-print-intf

### LINKED ISSUE #3230: Remove corebuild invocations in ffi chapter
See #3122 

The only two disabled test in the ffi chapter are disabled because `corebuild` doesn't work in the duniverse. We should replace those by dune invocations

NathanReb:
Fixes #3230


NathanReb:
It's an omission, it shouldn't be here. I'll remove it!

avsm:
Awesome! Farewell ocamlbuild!

Issue #552, opened 2013-03-12
New comment on block [block-idp7593088]


heidihoward:
same issues at above

Issue #372, opened 2013-02-24
New comment on block [block-idp7593088]


gregtatcam:
This definition fails:

utop $ module Logon = struct
    type t =
      { session_id: string;
        time: Time.t;
        user: string;
        credentials: string;
      }
    with fields
Error: Parse error: "end" expected after [str_items](in [module_expr])                                     ─( 11:18:34 )─< command 10 >─────────────────────────────────────────────────────────────────{ counter: 0 }─utop $   end;;

PR #3291, opened 2020-04-03
[Records] Require ppx_jane and fieldslib for module Logon

shakthimaan:
Both `ppx_jane` and `fieldslib` are to required in Utop before defining module Logon in the Records chapter.

The following two closed issues are related:
* https://github.com/realworldocaml/book/issues/372
* https://github.com/realworldocaml/book/issues/552

The output of `dune runtest` is clean.

### DIFF
--- a/book/records/README.md
+++ b/book/records/README.md
@@ -439,6 +439,8 @@ it is standard practice to name the type associated with the module
 `t`. Using this style we would write:
 
 ```ocaml env=main2
+# #require "ppx_jane"
+# #require "fieldslib"
 # module Log_entry = struct
     type t =
       { session_id: string;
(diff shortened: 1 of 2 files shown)


yminsky:
Good catch.

This seems like a general problem, but this PR doesn't feel like quite the right way to fix it.  In particular, we have ppx_jane in the prelude for almost all directories.  That means that things work without requiring any explicit invocation of the syntax extensions, which seems like a mistake.

I'll try to mint some PRs to try to fix this in individual chapters, and see if I can clarify the text as well as the examples.

shakthimaan:
As I had mentioned earlier, I am reading the text in the chapters, and manually trying out the examples in a Utop session. If you would like the reader to use the provided code examples from the book, then I shall follow the same.

yminsky:
I've done the equivalent of this for a bunch of chapters, this one included, in this PR:

https://github.com/realworldocaml/book/pull/3313

I think that should resolve your issues here.  Do tell me if I'm missing something.

PR #3292, opened 2020-04-03
[Variants] Define Logon and Heartbeat before client_message

shakthimaan:
When running with Utop, both `Logon` and `Heartbeat` need to be defined for `env=logger` before defining client_message. Otherwise, it throws an `Error: Unbound module Logon`.

output of `dune runtest` is clean.


yminsky:
Yeah, that's a reasonable point, I have a different fix for this in #3325

PR #3293, opened 2020-04-03
[Variants] Remove user_sessions parenthesis for proper compilation

shakthimaan:
The definition of `messages_for_user` throws "Error: Syntax error: operator expected" for the first `as` in `~f:(fun (messages, user_sessions) as acc ...` line. The fix is to remove the parenthesis as seen in the patch.

Output of `dune runtest` is clean.


yminsky:
I see no such syntax error in our build.  CI passes on the tip, so I'm pretty sure there's no syntax error.  Are you maybe using a different version of OCaml? Though I don't think the syntax rules around this have changed either.

Can you say more about how you encountered the error?  Was dune runtest not working for you?

shakthimaan:
The `dune runtest` did run fine, but, the error showed up when trying out the example manually in an Utop session. I am using utop version 2.4.3 with OCaml version 4.09.0.

yminsky:
I can't reproduce with utop 2.5.0 and OCaml 4.09.0.  Closing, but please reopen if you figure out why you're seeing this problem but other reproductions don't.

PR #3295, opened 2020-04-14
Prototype for a new example workflow

NathanReb:
This PR sketches a new way to write examples. For now only the "Files, Modules and Programs" chapter has been ported.

Each individual example has to live in its own directory. Examples are split in two subdirectories, `correct/` and `erroneous/`. A script will generate dune aliases to be able to build each example as a standalone project. By building here we mean running `dune build @all`. The name of the alias is the name of the example directory in `correct/`.

These alias are also registered as dependencies of the `runtest` alias so that when we run `dune runtest` each valid example is built.

This already allowed me to find an outdated example!


NathanReb:
You'll not that the example I mentioned above is still broken. I'm happy to update the code and the chapter but we just need to find a new list function to replace `intersperce` which is provided by `base` nowadays!

NathanReb:
If the proposed solution seems to be a good fit I'll start updating the other chapters. This takes a bit of time and is hard to automate as one has to review each example individually to decide how to best port them to this new scheme. Some of the existing examples aren't necessarily structured as a standalone dune project.

NathanReb:
Note to self: I need to make the build rule `dune build @all @runtest` before we merge this otherwise it most likely won't work with the testing chapter!

yminsky:
I checked this out and had trouble building it.

```
+|File "/home/yminsky/Documents/code/rwo/_build/install/default/lib/stdio/dune-package", line 1, characters 11-14:
+|1 | (lang dune 2.5)
+|               ^^^
+|Error: Version 2.5 of dune is not supported.
+|Supported versions:
+|- 1.0 to 1.12
+|- 2.0 to 2.4
```

Perhaps related , the repo has version 2.4.0 of Dune in the Duniverse, but I'm running 2.5.0 from the outside: 

```
[yminsky@viserion rwo]$ dune --version
2.5.0
```

yminsky:
I've now I've fixed up some bits (mostly, the example that was flagged with "intersperse"), and looked over the new setup.  Once the build passes, I'm happy to merge this.

NathanReb:
Great!

What was the reason for the dune error in the end?

yminsky:
I don't actually know!  In the end, I blew away my old switch (I'm using local opam switches), and then had to do a dune clean, at which point, things worked again.  Confusing.  I have no idea why I got the permission errors along the way.  Feels like there's some dependency bug somewhere.

NathanReb:
Do you have any idea where those CI errors are coming from? It appears we're getting `gcc` errors for ctypes. CC @avsm

avsm:
The CI error was fixed in #3302 as it was a latent bug in how we had the conf-ncurses package in opam-repo. Merging!

PR #3298, opened 2020-04-15
update duniverse to eliminate jbuild

avsm:
All of the jbuild uses are now gone, thanks to @rgrinberg
releasing a new atdgen!  This in turn means that a
duniverse pull now cleanly works without needing to run
a dune upgrade.


avsm:
The warning from ppxlib will disappear in a future PR thanks to https://github.com/ocaml/opam-repository/pull/16205

PR #3302, opened 2020-04-20
update to dune 2.5.1 and fix ncurses depexts

avsm:


yminsky:
Urk.  This seems basically unreviewable through github.  Too many files changed!  That's why it would be nice to get a version of this where we don't literally include all the code when we bump the duniverse.

Anyway, I guess you should merge it based on your knowledge of how you constructed it.  I don't see what else to do.

avsm:
Yeah. I'm working on the non-checked out duniverse next (it depends on having a `duniverse` binary available via homebrew to avoid bootstrap issues, but that isn't far off now).

For now, ocaml-ci is passing, so merging!

avsm:
Oops, waiting for Travis to be sure actually.

PR #3303, opened 2020-04-22
[First-Class Modules] Changing local variable from `Bumpable` to `Bumper`

aPere3:
In the _First-class modules_ chapter, I've had a hard time understanding this line:
```ocaml
# let (module Bumpable) = int_bumper in Bumpable.bump 3
```

There are two reasons for that:
+ Most importantly the name `Bumpable` is already used as a module type earlier. Here it is not used as such, but as a mere local variable introduced by the pattern matching.
+ The inlining of the `let .. in ..` syntax, makes it hard to differentiate with a usual `let` binding, given the previous reason.

Motivated by peoples on the discord channel, I propose to make two changes to this chapter:
+ Turn the first `let .. in ..` construct to a multi-line with a let binding to make the structure clearer.
+ Use `Bumper` as local variable name rather than `Bumpable` where suited.


yminsky:
This is a good fix.  I did a somewhat more extended version of the change in this commit:

https://github.com/realworldocaml/book/commit/f1ebdf7e382ca6211676b5071bb5dc1af3be2a9c

Thanks for noticing the issue!

PR #3308, opened 2020-04-29
[Imperative Programming] Fix missing Open Poly

g1eny0ung:
During reading the chapter of `imperative programming`, in section `Example: Doubly Linked Lists`, the code example in my local environment can't build as expected:

> let is_empty t = !t = None
Error: This expression has type 'a option but an expression was expected of type int

After searching, I found that the base library only exposes integer comparison by default. So, I add `open Poly` in:

```ocaml
let is_empty t = let open Poly in !t = None
```

Then the error disappeared.


yminsky:
I'd rather fix this by avoiding polymorphic compare, rather than reintroducing it.  (You can see the section on the perils of polymorphic compare in this chapter for the reasoning: https://dev.realworldocaml.org/maps-and-hashtables.html)

Using Option.is_none instead should take care of it.  Would you mind approaching it that way instead?

g1eny0ung:
I haven't read this chapter yet but I believe your approach is correct through your description.

```ocaml
let is_empty t = if Option.is_none !t then true else false
```

Amend it like this? If it is correct, I will update the PR.

yminsky:
That will work, but even better would be:

```ocaml
let is_empty t = Option.is_none !t
```

g1eny0ung:
Thx! Already updated.

PR #3309, opened 2020-04-29
Port Error Handling chapter to new example workflow

NathanReb:
This enables automatic checking of the examples in the Error handling chapter. There are a couple things to note here:
1. I removed the `result.ml{,i}` files from the example folders. They were used to add Ocaml interface blocks, like the examples from the Variables and Functions chapter and were therefore unchecked. I removed the example file and disabled MDX checks on the block until we can have ocaml interface blocks in MDX.
2. One of the examples is checked in an mdx block to showcase a runtime error but I marked it as correct since we're always supposed to be able to build anything in it. It adds a bit of redundancy but at least we have an alias if we want to build that specific example.
3. The last example contains benchmarks. I also put it in the `correct/` folder even though it's checked by running the executable in the chapter's shell blocks. Again a bit of redundancy but it can't hurt.

You'll also note that I changed a `non-deterministic` label from `command` to `output` so that the `dune exec` is actually run. I'm not even sure how it worked before since that dune exec is required for the subsequent block to work since it depends on the executable being built.


NathanReb:
After a quick investigation, it seems that the `non-deterministic` behaviour is broken and the commands are never run. I'll look into it.

This PR is still an improvement though so that shouldn't prevent us from merging it!

PR #3310, opened 2020-05-01
update duniverse to latest snapshot

avsm:


avsm:
We'll soon not need to check the vendored sources in, but for now just doing it in place.

PR #3312, opened 2020-05-02
tiny autogenerated change to octavius file. Not sure why this occurs...

yminsky:
This seems to be auto-deleted by dune build.  @avsm, do you know why?


avsm:
Hm it's a minor bug in dune -- it varies the version generation based on the build context.  I'll just dig in a bit further before merging this.

What we could also use is a release of ppx_jane_style that uses odoc.parser, so that we can ditch the ancient octavius dependency entirely...

PR #3313, opened 2020-05-04
Remove ppx jane from prelude

yminsky:
Remove ppx_jane and open Base from a bunch of prelude.ml files.  This required a bunch of follow-on changes to merge env's together, and to add some #require's to explicitly turn on syntax extensions.  This all seems good.

There are some weird corners, notably, for some reason ppx_sexp_conv doesn't seem to work when #require'd, but ppx_jane does.  I posted on issue there, and will fix this when that is resolved:

https://github.com/janestreet/ppx_sexp_conv/issues/30


yminsky:
Note that there's a bunch of whitespace-only changes here that for some reason Github is incompetent at hiding, even when you turn on whitespace hiding.  If you look at the diff in Patdiff, everything looks fine.

avsm:
@NathanReb there's a strange anomaly in environment variable handling.

If I run this PR with

```
env TZ=UTC dune runtest
```

...it passes the expect test.

However, forcing the env-var in either dune-workspace with:

```
(lang dune 2.0)
(env (_ (env-vars (TZ UTC))))
```

fails to force it, and the expect tests fail.

The minimal place to force the env-var to be set would be in the `dune` file for the README.md, but the `mdx` invocation doesn't allow an environment variable to be set. Would you be able to investigate this one?

NathanReb:
Have you tried setting it with MDX labels? It should work and might be a good work around until we figured this out!

NathanReb:
I.e. adding `,set-TZ=UTC` to the block opening line should work. There's a high chance these labels are ignored for toplevel blocks but if so, that's a bug and we should fix it!

yminsky:
For now, I've added non-determinism markers.  We can remove them when we have a solution.

PR #3314, opened 2020-05-05
Testing chapter with example workflow

NathanReb:
This PR was built on top of https://github.com/realworldocaml/book/tree/testing to bring it up to speed with our latest tools and practices.

Given that the chapter is incomplete I removed it from the generated web and PDF versions but the chapter sources are there and will be checked by the CI. It will be easier to keep the chapter up to date if we merge this. Once it is complete we can revert the last commit to include it in the book.

Let me know what you think!


NathanReb:
I'm a bit puzzled by the CI failures. It seems that the build for the `manual_property_test` example fails but the logs aren't very helpful as to why it is failing or how and I can't reproduce it locally where the tests pass for the chapter. I'll take a deeper look!

If one of you manages to reproduce it locally in the meantime, please let me know!

NathanReb:
Ah nevermind, I found it. The example dependencies weren't properly specified and my local build was succeeding because of a race condition!

yminsky:
What's currently blocking us from releasing this as a WIP? i.e., in the tree, but not on the site by default, and with some way of building a "full" version that includes the chapter.

NathanReb:
Hmm I'm suprised CI went green I'll try to see if the race condition is still around so we can merge this ASAP!

NathanReb:
I still get failures locally from a clean build, I'll try to better specify the dependencies.

NathanReb:
I just fixed the examples dependencies and got rid of the race conditions, this should be ready to merge!

yminsky:
Does this have a way of suppressing this chapter from being published?  If so, could we make sure the workflow for WIP chapters is documented?

NathanReb:
Good idea, I'll document what I did in this commit: https://github.com/realworldocaml/book/pull/3314/commits/5b9e1049ce8998c4666e7e1611b1a08cafb31c15

There's no particular mechanism for that though, I just removed it from a few dune rules and the `toc.scm` files so that we don't generate the HTML nor include it in the PDF. There might be a better way to do this, in particular some of the rules might be generated and we could keep a black or whitelist of chapters for the book. That would make it easier for authors of WIP chapters when they want to enable/disable the generation of those files.

yminsky:
Hmm.  I do want something else, which is a target I can build to generate the full site.  e.g., @wipsite, or something like that?

NathanReb:
I just rebased and will merge as soon as travis goes green!

NathanReb:
Same as in #2883, github didn't pick the travis build here: https://travis-ci.org/github/realworldocaml/book/builds/699341195.

PR #3317, opened 2020-05-06
Port Imperative Programming chapter to new examples workflow

NathanReb:
This PR ports the Imperative Programming chapter's examples.

1. I got rid of the examples that weren't actual examples and made them skipped blocks instead (e.g. `remember_type.ml` and `let_rec.ml`).
2. I removed the unused `numbers.txt`
3. I extracted the `dictionary` example. You'll note that I merged the `dictionary2.ml` into it, as an extra function. The blocks in the chapter remain unchanged but the code is now checked to see if it compiles for both. Doing so I discovered that it was relying on polymorphic comparison operators. I added a couple local `Poly` open since the arguments are supposed to be polymorphic but I'm happy to replace it with a better solution if there's one.
4. I extracted the `dlist` example
5. I reworked the `time_converter` example a bit and added a test to ensure that if it was executed as in the chapter's example it wouldn't crash.
6. I extracted a `time_converter2` example which is basically a copy of `time_converter` but with the alternative, printf-based implementation.

I also took the liberty to edit the outdated text about the weakly polymorphic types as it was out of sync with the OCaml version we use.

Let me know what you think!


yminsky:
I made a small change to the text, but otherwise, looks great.

Issue #3340, opened 2020-06-04
symbols of infix operators is not accurate


arbipher:
In the section Prefix and Infix Operators of chapter Variables and Functions, it says

> A function is treated syntactically as an operator if the name of that function is chosen from one of a specialized set of identifiers. This set includes identifiers that are sequences of characters from the following set:
> ! $ % & * + - . / : < = > ? @ ^ | ~

It's not accurate on `! ~ ?` according to the [document](https://caml.inria.fr/pub/docs/manual-ocaml/lex.html#sss:lex-ops-symbols) (thanks to Christophe in OCaml discord). They cannot be the first char for `core-operator-char`

yminsky:
Thanks for the catch!  If you're up for posting a PR to fix it, that will speed things up, but if not, I'll get this fixed at some point.

PR #3341, opened 2020-06-05
symbols of infix operators is not accurate

arbipher:
It's for #3340


yminsky:
Merged, with some small tweaks. Thanks!

PR #3342, opened 2020-06-09
Added function call example "Guided Tour"

AndrewCarterUK:
The description in the "OCaml as a Calculator" section explains that function arguments are separated by spaces instead of by parentheses and commas, but the example does not include a case that demonstrates this.

This PR adds an example of using the `Float.max` function on two numbers.

### DIFF
--- a/book/guided-tour/README.md
+++ b/book/guided-tour/README.md
@@ -77,6 +77,8 @@ Now let's try a few simple numerical calculations:
 - : int = 100
 # 3 * 5 > 14
 - : bool = true
+# Float.max 2.7182 3.1415
+- : float = 3.1415
 ```
 
 By and large, this is pretty similar to what you'd find in any programming


AndrewCarterUK:
Just read a bit further along and realised that what I have done has contradicted slightly with a later paragraph:

image

I'm not sure what the best approach would be here? It seems strange to refer to an example that doesn't exist, but as this PR stands it breaks the flow of the tutorial somewhat.

Another option would be to migrate the wording about the function syntax down into the "Functions and Type Inference" section.

I'm happy to close the PR and open an issue for discussion - or - adjust this PR following some direction as to what is the best approach?

yminsky:
How about this as a solution: 

https://github.com/realworldocaml/book/pull/3345

AndrewCarterUK:
Closing in favour of #3345

PR #3344, opened 2020-06-11
Modify Lists & Patterns chapter filenames extensions example

jplew:
At the risk of sounding pedantic, this example in the Lists & Patterns chapter fails to handle some file-naming edge cases:
 
```ocaml
let extensions filenames =
  List.filter_map filenames ~f:(fun fname ->
      match String.rsplit2 ~on:'.' fname with
      | None  | Some ("",_) -> None
      | Some (_,ext) ->
        Some ext)
  |> List.dedup_and_sort ~compare:String.compare
;;
```

If I extend the list of filenames from `["foo.c"; "foo.ml"; "bar.ml"; "bar.mli"]` to `["foo.c"; "foo.ml"; "bar.ml"; "bar.mli"; "baz"; ".baz"; "baz."]`, I obtain the following:

*Before*
```
utop # extensions ["foo.c"; "foo.ml"; "bar.ml"; "bar.mli"; "baz"; ".baz"; "baz."];;
- : string list = [""; "c"; "ml"; "mli"]
```

*After*
```
utop # extensions ["foo.c"; "foo.ml"; "bar.ml"; "bar.mli"; "baz"; ".baz"; "baz."];;
- : string list = ["baz"; "c"; "ml"; "mli"]
```

Perhaps it's a matter of taste, but I believe the latter is the desired behaviour.

### DIFF
--- a/book/lists-and-patterns/README.md
+++ b/book/lists-and-patterns/README.md
@@ -52,6 +52,7 @@ the box containing `3`) points to the empty list.[lists/structure of]{.idx}
   <img src="images/lists-and-patterns/lists_layout.png"/>
 </figure>
 
+
 Each `::` essentially adds a new block to the proceding picture. Such a block
 contains two things: a reference to the data in that list element, and a
 reference to the remainder of the list. This is why `::` can extend a list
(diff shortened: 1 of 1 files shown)


yminsky:
Yeah, it really is a matter of taste, and I prefer the semantics of what's currently there.  It's certainly behaving in the way I expected it to...

PR #3345, opened 2020-06-11
move description of how multi-argument functions work

yminsky:


AndrewCarterUK:
I think this is a good change :+1:

PR #3346, opened 2020-06-11
Add missing commas

bts:
Hi there,

Thanks for the great book! I added some commas where they were needed:

> libraries tools and techniques

but also made a judgement call and suggested ", as well as" instead of the first "and". If you'd like, let me know and I can change that back to ", and".

Thanks!
Brian


yminsky:
Made some tweaks and merged.

PR #3349, opened 2020-06-17
Add missing dependencies to dictionary example

NathanReb:


yminsky:
The change looks reasonable, but: why is CI not passing now? And why did CI pass before this feature, if these dependencies are in fact necessary?

NathanReb:
CI is passing: https://travis-ci.com/github/NathanReb/book/builds/171858069, ocaml-ci is broken for us atm the moment, the failure is irrelevant.

For some reason github failed to pick up the travis builds and isn't reporting their status, I'll push an empty commit to see if I can re-trigger it.

NathanReb:
The CI was green before because some of the dependencies were built as dependencies of other rules. It's not great that dune lets this slip through but unfortunately the `(package ...)` dependencies aren't very reliable with regards to that.

Only way to reliably spot this  kind of badly specified dependencies is to try building just the example after a `dune clean`. I ran into this one by accident.

PR #3352, opened 2020-06-27
Fix cli examples (WIP)

yminsky:


yminsky:
I imagine I'm doing several things wrong.  Some weird things about this PR:

- Running "make dep" updates a bunch of seemingly unrelated stuff
- the dune.inc file is empty, and I don't know how to get the correct contents generated.
- Despite having followed some of the instructions, things don't quite seem to work.  I only added one of the .rwo-example files, so maybe I need to add the rest first?  But right now, when I do a dune runtest, it looks like everything passes, even though I broke an example to make sure it didn't.

book/dune
```diff
@@ -1,12 +1,305 @@
+(alias
```
NathanReb, line 1: This looks like a rebase/merge mishap, `book/dune` shouldn't be impacted by your changes and in particular here it seems to revert some of the work we've done on the HTML and LaTeX rules.

NathanReb:
Indeed for the examples to be built you need to promote the generated rules to the `dune.inc` file. Running `dune runtest` should fail and show you the diff with the new rules. From there running `dune promote` and `dune runtest` again should pick up the error.

I'm surprised it didn't show up, from a quick look the dune file looks right. If you want to just run the rule generation bit you can run `dune build @@book/command-line-interface/examples/runtest`. The double `@` makes it non recursive and only executes actions attached to the runtest alias in this particular directory. For us that's just the rule generation. 

As to the `make dep`, I think this is just a legacy target that we don't use anymore, we can probably remove it.

book/command-line-parsing/README.md
```diff
@@ -215,11 +215,11 @@ And that's all it took to build our little MD5 utility! Here's a complete
 version of the example we just walked through, made slightly more succinct by
 removing intermediate variables.
 
-```ocaml file=examples/md5_succinct/md5.ml
+```ocaml file=examples/correct/md5_succinct/md5.ml
 open Core
 
 let do_hash file =
-  Md5.digest_file_blocking file
+  Md5.digest_file_blocking filee
```
avsm, line 222: `filee` -> typo?

yminsky: Not a typo.  I broke one of the examples to see if the test would break, and it didn't.  I'm still trying to get this fixed.

yminsky:
@NathanReb Nathan, did you do a PR to fix the documentation for how to set up the new examples workflow? I believe there was some extra file I needed to delete, but I don't remember where.  Anyway, it would be good to have the README up to date.

yminsky:
OK, this now all works, I believe.  I'll merge it once CI passes.

PR #3353, opened 2020-06-29
Remove rwo-dep from the codebase

NathanReb:
It's not used anymore as pointed out in #3352 !


NathanReb:
Here's the successful build that github failed to pick, again: https://travis-ci.org/github/realworldocaml/book/builds/703177580

Merging!

PR #3354, opened 2020-06-30
Allow marking chapters as WIP and add extra targets for website and PDF including them

### LINKED ISSUE #3348: Add targets to generate website and PDF with WIP chapters

NathanReb:
Fixes #3348

This PR adds a way to mark a chapter as Work In Progress in the table of content file `toc.scm` along with two new targets, `@site-wip` and `@pdf-wip`.

The regular website and PDF targets will generate them, excluding the WIP chapters while the `-wip` targets will include everything.

The `-wip` targets are built to a separate `static-wip` folder.

For now some of the rules are duplicated between `static/dune` and `static-wip/dune`, with few modifications. We could eventually generate both to make sure one is not lagging behind by accident but my impression is that for now these rules aren't likely to change too often.


NathanReb:
I just need to document this in the README but then it should be good to go.

bin/lib/book.ml
```diff
@@ -281,28 +295,28 @@ type src = [
 | `Latex
 ]
 
-let make ?(repo_root=".") ~out_dir = function
+let make ?(repo_root=".") ?(include_wip=false) ~out_dir = function
```
yminsky, line 298: Is there a reason for this argument to be optional?  I suspect this isn't called in enough places for the concision to be worth making the choice implicit.

NathanReb: I made it optional because for the FAQ and the Install pages the argument is unused but I'll make it mandatory!

NathanReb: Hmm now that I look at it it feels weird to make it mandatory and to pass a value in `bin/bin/app.ml` knowing it won't be used.

yminsky:
What's blocking this from being released?

bin/lib/book.ml
yminsky: I don't really feel strongly. I'm happy with it being released as is.

NathanReb:
Nothing beside that usual travis/github issue

NathanReb:
Actually there's something that needs fixing and hopefully that new build will be picked up by the github integration this time.

PR #3355, opened 2020-07-04
Fix examples in more chapters

yminsky:
This fixes records, variants, and async.


NathanReb:
I think each individual example lacks a proper `.rwo-example`! This can work due to race conditions when building the whole repo but trying to build those examples from scratch will fail.

yminsky:
Oh, interesting. I assumed it was OK because there was a standard set of libraries there by default, or something like that.  But I guess that was just wrong.  I'll fix this in a separate PR.

PR #3367, opened 2020-08-02
Update json examples

yminsky:
This still doesn't build, in part because we're missing a dependency (shell)


avsm:
Finished up the JSON chapter and also updated to core v0.14 from v0.13 while adding the shell package.

avsm:
Fixing up that menhir error here https://github.com/dune-universe/opam-overlays/pull/66 -- its because of a divergence in the source archives and the git repo upstream

PR #3374, opened 2020-09-28
Unused endpoint1 and endpoint2 in Segment case

yetkinozturk:
matching case `Segment { endpoint1; endpoint2 }` and returning false raises compilation warnings/errors for unused variables.  
Using _ makes Segment definition quite idle but not it compiles fine.

### DIFF
--- a/book/guided-tour/README.md
+++ b/book/guided-tour/README.md
@@ -991,7 +991,7 @@ tell *utop* to process the input, not to separate two declarations
     | Rect { lower_left; width; height } ->
       point.x    > lower_left.x && point.x < lower_left.x + width
       && point.y > lower_left.y && point.y < lower_left.y + height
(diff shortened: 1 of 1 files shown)


book/guided-tour/README.md
```diff
@@ -991,7 +991,7 @@ tell *utop* to process the input, not to separate two declarations
     | Rect { lower_left; width; height } ->
       point.x    > lower_left.x && point.x < lower_left.x + width
       && point.y > lower_left.y && point.y < lower_left.y + height
-    | Segment { endpoint1; endpoint2 } -> false
```
yminsky, line 994: A catch-all case is not good style.  Changing this to `Segment _ -> false` would be better.

yetkinozturk: updated.

PR #3395, opened 2020-10-24
Fix online list docs broken link

colinchan15:
The "online docs" link in the [`More useful list functions`](https://dev.realworldocaml.org/lists-and-patterns.html#more-useful-list-functions) of the `Lists and patterns` chapter leads to a [broken link](http://realworldocaml.org/doc). 

Technically, it should lead to: https://ocaml.janestreet.com/ocaml-core/latest/doc/, but since the section is on the `List` functions from `Base`, I believe it should lead to: https://ocaml.janestreet.com/ocaml-core/latest/doc/base/Base/List/index.html


yminsky:
Thanks! That's the right fix here. I should also find other examples of the dead links, which I'm sure show up in other places.

PR #3397, opened 2020-10-25
Fix list-and-patterns section reference to List.dedup_and_sort

colinchan15:
In the [FILTERING WITH LIST.FILTER AND LIST.FILTER_MAP](https://dev.realworldocaml.org/lists-and-patterns.html#filtering-with-list.filter-and-list.filter_map) section, the code example references the `List.dedup_and_sort` function while the text references a `List.dedup` function which does not exist in the `List` module of `Base`. I am making the text consistent with the code example by changing:

`... List.dedup to remove duplicates.` -> `... List.dedup_and_sort to return the list with duplicates removed and in sorted order.`


yminsky:
Thanks for the catch!

PR #3398, opened 2020-10-27
WIP - PDF utop code formatting more consistent with HTML

### LINKED ISSUE #3275: Semi-colons not apparent in generated PDF
Hi.

Compared to the generated HTML version of this book, the PDF version found at http://dev.realworldocaml.org/book.pdf lacks the `;;` terminating semi-colons for all examples where it is used with `utop`.

Just wanted to pass this along. Thanks!

srinchiera:
EDIT - I realized that this breaks multi-line comment syntax highlighting. Will fix

This change makes PDF code formatting more consistent with HTML code formatting in 2 ways:
1. All lines in a top-level utop expression are prefixed with `#`. Previously, only the first line was prefixed with `#`.
2. All top-level utop expressions will end with `;;`. If the top-level expression is a single line, then the double semicolon will appear in the same line as the expression. If the top-level expression is multi-line, then the double semicolon will be appended on a new line.

Photos that illustrate updated formatting:

Single line:
* Previous PDF
PrevPdf2
* New PDF
NewPdf2
* Current HTML
CurrHtml2

Multi line:
* Previous PDF
PrevPdf
* New PDF
NewPdf
* Current HTML
CurrHtml

Fixes #3275 

I originally noticed this discrepancy after reading text that hinted that a double-semicolon might be missing: https://github.com/realworldocaml/book/blame/master/book/guided-tour/README.md#L818

An annoying part about code formatting in the PDF is that, unlike the HTML, a user cannot select top-level utop expressions without also selecting the `#` characters. I wonder if there is a way to configure the `listings` package so that mouse selection behaves more like the HTML.

### DIFF
--- a/bin/bin/convert_md.ml
+++ b/bin/bin/convert_md.ml
@@ -14,17 +14,23 @@ let pp_html ppf s =
   in
   String.iter add s
 
-let pp_output ppf = function
+let pp_output_html ppf = function
   |`Output s -> Fmt.pf ppf ">%a\n" pp_html s
   |`Ellipsis -> Fmt.pf ppf "...\n"
 
- let pp_line ppf l = Fmt.pf ppf "%a\n" pp_html l
+let pp_output_latex ppf = function
+  |`Output s -> Fmt.pf ppf "%s\n" s
+  |`Ellipsis -> Fmt.pf ppf "...\n"
 
- let pp_toplevel ppf (t:Mdx.Toplevel.t) =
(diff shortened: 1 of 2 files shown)


yminsky:
Looking back at this: does this now fix the multi-line comment syntax highlighting? Is it ready for review?

srinchiera:
> Looking back at this: does this now fix the multi-line comment syntax highlighting? Is it ready for review?

This is not yet complete. I will close out the PR for now, and re-open when there is a fix.

PR #3399, opened 2020-10-28
Remove empty code block in First-Class Modules chapter

srinchiera:


book/first-class-modules/README.md
```diff
@@ -822,11 +822,6 @@ Now build this into a command-line interface to experiment with it:
   (preprocess (pps ppx_sexp_conv)))
 ```
 
-
```
yminsky, line 825: While we're at it, can you remove the above scheme block as well?

avsm:
Fixed in #3435

PR #3403, opened 2020-11-09
Set a fixed dune version to use in RWO

NathanReb:
Our current set of rules for code examples only work if the opam installed dune has the same version as the vendored `dune-configurator` and `dune-build-info` have.

This PR express that constraint and updates the duniverse accordingly.


yminsky:
Hmm.  This doesn't build, and is a little hard to read. Also, I expected 2.6.2 to be gone, but it still seems like it's there?  I'm surprised that the committed version hasn't moved to 2.7.1.

emillon:
This should be made unnecessary by #3405.

PR #3404, opened 2020-11-12
Update README.md

arbipher:
The difference between `Base.Comparator` and `Core.Comparator` is not well-documented.
The silent shadowing is confusing sometimes due to the unclear error message.
I am thinking many people like me use this book as the document and tutorial for Base and Core.
Therefore, I suggest mentioning this in this section though the book may focus on Base.
In fact, even the official document contains errors on `sexp_of` ([link to doc](https://ocaml.janestreet.com/ocaml-core/latest/doc/core_kernel/Core_kernel/Map/index.html) and [my pull to fix there](https://github.com/janestreet/core_kernel/pull/102))

### DIFF
--- a/book/maps-and-hashtables/README.md
+++ b/book/maps-and-hashtables/README.md
@@ -642,6 +642,49 @@ you can always write your own comparison function by hand; but if all you
 need is a total order suitable for creating maps and sets with, then
 `[@@deriving compare]` is a good choice.
 
+If you use Core rather than Base, things are a bit different. `Comparator.Make`
+requires `t_of_sexp` for the generated serialization modules. For this, we change
+the ppx from `sexp_of` to `sexp`.
+
+```ocaml env=main
+# #require "ppx_jane"
+# module Book = struct
+    module T = struct
+      type t = { title: string; isbn: string }
+      [@@deriving compare, sexp]
+    end
+    include T
+    include Comparator.Make(T)
+  end
+module Foo :
+  sig
+    module T :
+      sig
+        type t = { title : string; isbn : string; }
+        val t_of_sexp : Sexp.t -> t
+        val sexp_of_t : t -> Sexp.t
+        val compare : t -> t -> int
+      end
+    type t = T.t = { title : string; isbn : string; }
+    val t_of_sexp : Sexp.t -> t
+    val sexp_of_t : t -> Sexp.t
+    ...
+    module Replace_polymorphic_compare : ...
+    module Map : ...
+    module Set : ...
+```
+
+The module `Foo` has inner modules `Map` and `Set`. Theses modules are for 
+serialization and they are not compatible with `Core.Map` or `Core.Set`.
+When you open this module locally, the `Core.Set` will be shadowed by 
+`Foo.Set` silently. 
+
+```ocaml env=main
+# let play_foo foo =
+    let open Foo in
+    Set.add ...
+```
+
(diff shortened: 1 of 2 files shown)


yminsky:
Checks on this have failed, presumably because you didn't do a "dune runtest" followed by a "dune promote". Can you try to fix it up?

arbipher:
I see. Let me fix it and will make a new commit then.

arbipher:
OK, I pushed a new commit. I use a top-level code snippet (`ocaml env=core_compare`) to show what `Core.Comparator` derives, and an example file to show `Core.Comparable` generate inner Set and Map modules. I didn't find how to suppress the _generated code_ for top-level code. `Comparable.Make(T)` is too long.

However, it's not perfect since top-level code can show the auto-generated error message but now I can only put the expected error message in the file comment.

arbipher:
I thought `Core.Set` and the generated `Book.Set` were always not compatible. My personal code reported errors on this. That's the start point where I check the combination of {Base, Core} * {Comparator, Comparable}. However, I just found this code works

```ocaml 
# #require "ppx_jane"
# open Core
# module Book = struct
  module T = struct
    type t = { title: string; isbn: string }
    [@@deriving compare, sexp]
  end
  include T
  include Comparable.Make(T)
end

 # Book.Set.empty;;
- : Book.Set.t = <abstr>

 # Set.empty (module Book);;
- : (Book.t, Book.comparator_witness) Set.t = <abstr>

# Set.add (Book.Set.empty) {title="RWC"; isbn="12345"};;
- : (Book.t, Book.comparator_witness) Set.t = <abstr>

# Book.Set.add (Set.empty (module Book)) {title="RWC"; isbn="12345"};;
- : Book.Set.t = <abstr>
```

I remove the word saying there are not compatible (but I still wonder they should not be compatible).

As for this pull request, `sexp_of` and shadowing global `Set` are still worthing to point out, even these sets are compatible.

book/maps-and-hashtables/README.md
```diff
@@ -642,6 +642,68 @@ you can always write your own comparison function by hand; but if all you
...
+    type t = T.t = { title : string; isbn : string; }
+    val compare : t -> t -> int
+    val t_of_sexp : Sexp.t -> t
+    val sexp_of_t : t -> Sexp.t
+    type comparator_witness = Base.Comparator.Make(T).comparator_witness
+    val comparator : (t, comparator_witness) Comparator.t
+  end
+```
+
+`Core.Comparable` generates inner modules `Map` and `Set` for serialization. When you open this module locally, the `Core.Set` will be shadowed by
```
yminsky, line 678: Hmm.  I'm not excited about this way of talking about it, because the real mistake in the code here is the local open of Book. Generally speaking, opening modules should be minimized, and there's really no justification for opening Book here.

I agree that it would be good to have a little subsection explaining the differences between Base and Core/Core_kernel, but I don't think this is quite getting at it.  I think I'll take a separate swing at this in a new PR.

arbipher: It's subtle here but I agree that the code is not expressive enough. I read the [open modules section](https://dev.realworldocaml.org/files-modules-and-programs.html#opening-modules) and the intended scenario is like `print_median`, e.g. `collect_median` where you add the median to another Set or Map. I will back on this when I have a good example for this.

I will make a separate PR, just explaining the differences in short words later.

PR #3405, opened 2020-11-12
Example rules: pick dune in PATH

emillon:
Previously, example rules would invoke dune using `(run dune ...)`.

When seeing `(run prog ...)`, dune first looks for a public executable named `prog` in the workspace before looking at the environment (`PATH`).

In the case of this repository, it happens that a dune executable is in the workspace, because the sources of dune-build-info are there and they are the same as the dune sources.

When building the examples, this dune (2.6 at the moment) will be picked, and it will try to read dune-package files for the other packages. If these have been written by a dune in the environment with a higher version, it will fail with an error. Effectively, this puts a constraint that dune in the environment should be older than the vendored one.

To fix this, the `(run dune ...)` action is replaced by `(system "dune ...")`. This is treated as an opaque string passed to the system shell, which resolves using the environment only. This causes the vendored dune to be ignored.

Thanks @NathanReb for the idea!

### DIFF
--- a/bin/examples-rules/examples_rules.ml
+++ b/bin/examples-rules/examples_rules.ml
@@ -43,7 +43,7 @@ let print_rule ~dir_name ~path (config : Config.t) =
  (deps
   (source_tree %s)%s)
  (action
-  (run dune build @all @runtest --root %s)))
+  (system "dune build @all @runtest --root %s")))
 
 (alias
  (name runtest)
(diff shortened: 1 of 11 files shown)


emillon:
NB: this requires a `dune clean` to be done first. Otherwise there'll be a dune in `_build/install/default/bin` which will be picked up.

avsm:
Thanks!

PR #3413, opened 2020-11-27
Update README.md

jakeknigge:
...just a minor typographical nit.


avsm:
I think the existing text is correct: "The previous example we worked through // touched on". Thanks for the suggestion though!

avsm:
Having said that, @yminsky picked up your suggested text in https://github.com/realworldocaml/book/pull/3414 :-)

PR #3412, opened 2020-11-24
Switch to opam-monorepo

NathanReb:
Fix #3407 

I'm marking this as draft until I updated the README with a bit of documentation on how to work with opam-monorepo.

@emillon can use this to test the ocaml-ci counterpart to this change!

PR #3417, opened 2020-12-03
Build site and pdf by default

### LINKED ISSUE #3416: HTML and latex generation are not checked by dune build @install
In #3406, some blocks look like this:

    ```sh skip
    >>> (ls .)
    Could not find matching handler: ls
    ```

This triggers (and should) an error in mdx, but this only happens in `@html`, not in `@runtest`. As a result, it's missed by ocaml-ci which runs `@install` for monorepo builds. As a workaround, this alias (or `@site`? what about `@latex`?) can be added to `@install`.

emillon:
Closes #3416

### DIFF
--- a/dune
+++ b/dune
@@ -1,2 +1,8 @@
 (env
  (non-deterministic (env-vars (MDX_RUN_NON_DETERMINISTIC true))))
+
+(alias
+ (name install)
+ (deps
+  (alias_rec site)
+  (alias_rec pdf)))
--- a/dune-project
+++ b/dune-project
@@ -25,6 +25,7 @@ second edition book. See https://realworldocaml.org/")
   cmdliner
   cohttp-async
   conf-ncurses
+  conf-pandoc
   core
   core_bench
   ctypes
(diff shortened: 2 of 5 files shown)


emillon:
Building the html (the part that fails on the linked PR) requires pandoc, so we'll have to mark it as a depext. That will be easier to do once #3412 is ready.

emillon:
In terms of dependencies, this requires pandoc and texlive-full.

For pandoc, depending conf-pandoc works (the version is different from the one installed in Dockerfile, 2.9-1 from github vs 2.2.1-3 from debian, though I haven't compared the output).

For texlive it's a bit more involved: there's conf-texlive but it does not install all the required packages (and the build fails with a missing xcolor package). Adding texlive-full in depexts work. I suppose we could add a conf-texlive-full package which attempts to build a document that uses xcolor.

emillon:
Rebased and now green.

avsm:
I've got a bunch of PDF changes queued up for the new format we need to give the publisher, so I'll close this one for now.

PR #3420, opened 2020-12-11
Relock using latest packages

### LINKED ISSUE #3408: Upgrade to JS packages v0.14
We need to upgrade the suite of Jane Street packages from 0.13 to 0.14.

We should probably wait on #3407 so we can use this as a test run for the new upgrade workflow.

emillon:
Relock using latest packages

Closes #3408

This consists in the following commands:

  - (setup pins and repo as described in README.md)
  - `opam monorepo lock`
  - `opam monorepo pull`
  - `dune runtest`
  - `dune promote`


emillon:
A few workarounds are necessary to upgrade to the latest packages. I'll file the relevant bugs and document them here.

PR #3422, opened 2021-01-10
Fix typo in prologue

magurotuna:
Hi! I'm very new to OCaml. I found this book really beneficial to me as soon as I started reading it. Thanks for the awesome work of you all!

Anyway, when reading prologue, I noticed there was a tiny typo, so this PR fixes it :)


yminsky:
Thanks for the fix!

PR #3424, opened 2021-01-27
Fix references to Imperative Programming chapter.

freevoid:
References in two places were linking to the guided tour while it's
clear from the context that they should link to the dedicated chapter
about imperative programming.

References in two other places were titled "Imperative Programming 1",
in contrast with the rest titled simply "Imperative Programming".


yminsky:
Thanks for catching these. These look like artifacts from an earlier round of automated upgrades from the previous edition of the book.

PR #3427, opened 2021-06-29
Fix up some problems in the test chapter.

yminsky:
Mostly moving things into correct/erroneous directories, and adding .rwo-example files.


yminsky:
Weirdly, this still doesn't work for me locally.  When I build on my local box and just point the browser at, say, file:///home/yminsky/Documents/code/rwo/_build/default/static/guided-tour.html, the index file is not properly loaded.

This is what the console shows:

Screenshot from 2021-07-12 21-59-31

PR #3431, opened 2021-07-08
Optimise the GitHub Actions workflow

smorimoto:


.github/workflows/build.yml
```diff
@@ -16,7 +16,7 @@ jobs:
           - ubuntu-latest
           - macos-latest
         ocaml-compiler:
-          - 4.10.2
```
smorimoto, line 19: Use the latest compatible OCaml version at all times

.github/workflows/build.yml
```diff
@@ -28,30 +28,10 @@ jobs:
...
           ocaml-compiler: ${{ matrix.ocaml-compiler }}
+          dune-cache: ${{ matrix.os != 'macos-latest' }}
 
-      - name: Install Linux system packages
-        if: runner.os == 'Linux'
-        run: sudo apt-get -y update && sudo apt-get -y install libffi-dev libgmp-dev libpcre3-dev ncurses-dev pkg-config tzdata pandoc
-
-      - name: Install macOS system packages
-        if: runner.os == 'macOS'
-        run: brew install gmp libffi pcre pkg-config pandoc
```
smorimoto, line 38: Since Linuxbrew is installed on GitHub Actions by default, you can do it in one step

.github/workflows/build.yml
```diff
@@ -28,30 +28,10 @@ jobs:
...
-        uses: actions/cache@v2
-        id: cache
-        with:
-          path: |
-            _build
-            ~/.opam
-          key: dune-${{ runner.os }}-${{ matrix.ocaml-compiler }}-${{ github.sha }}
-          restore-keys: |
-            dune-${{ runner.os }}-${{ matrix.ocaml-compiler }}-${{ github.sha }}
-            dune-${{ runner.os }}-${{ matrix.ocaml-compiler }}-
```
smorimoto, line 50: The cache is no longer needed. It's much more reliable to use the dune cache instead

.github/workflows/build.yml
```diff
@@ -28,30 +28,10 @@ jobs:
...
-
-      - name: Install Dune
-        if: steps.cache.outputs.cache-hit != 'true'
-        run: opam install -y "dune>=2.9.0"
```
smorimoto, line 54: If dune cache is enabled, dune is automatically installed

smorimoto:
The dune cache is not available for the macOS runners because opam 2.0.9 is not yet released, but it shows that Ubuntu is fast enough. I believe opam 2.0.9 will be released next week, so we should be able to enable it very soon.
https://github.com/smorimoto/book/runs/3021546766?check_suite_focus=true

.github/workflows/build.yml
avsm: This one is a monorepo, so it actually has to be 4.10.2 specifically (for the expect tests).

avsm:
Very cool changes!

PR #3449, opened 2021-08-01
Fix typo

outkine:


yminsky:
this fails because you have to update the README.md file for that chapter to match the changed source code. If you do "dune runtest" and then "dune promote", that will do it for you, but it might not be worth getting the build working to do that!

yminsky:
Thanks!

PR #3470, opened 2021-09-26
update book to 4.13.0

avsm:


avsm:
ocaml-ci's expected to fail at this stage. I'm just checking the GHA run and then will update the lock files.

yminsky (approved):
Looks good to me. I didn't look at the duniverse pages, and the github UI is basically falling over because of the size of the diff.  But other than that, fine.

PR #3478, opened 2021-10-03
Part 3 edits

avsm:


book/compiler-backend/README.md
```diff
@@ -1040,24 +1020,20 @@ OCaml stack frames are too complex for Perf to understand directly, and so it
 needs the compiler to fall back to using the same conventions as C for
 function calls. On 64-bit Intel systems, this means that a special register
 known as the *frame pointer* is used to record function call history.
```
yminsky, line 1022: Perf these days can understand OCaml stack frames without frame pointers, though frame pointers make perf faster and the stack traces it gets more reliable

avsm: fixed in d7fb5c684c18f99f6de8f1d97f5925dd75464c88

PR #3484, opened 2021-10-06
Remove {.idx #fragment} syntax that causes render issue in PDF

kerrickstaley:
The syntax `{.idx #fragment}`, which appears in 2 places in the source markdown, causes text to appear in the PDF that shouldn't be there. Here are screenshots of the two places where it happens:

> 

> 

I'm not sure what this `{.idx #fragment}` syntax does, but the trend in recent commits has been to replace it with `{.idx}`. I did so in this PR and it fixes the PDF output. Hopefully this is the correct solution to the issue :)

> 

>


kerrickstaley:
I seem to have typoed my search or forgot to apply it to files other than `variables-and-functions/README.md`; this syntax actually occurs 42 times across all `.md` files in the repo. I'm guessing it's there for a reason and bulk removing it probably isn't the right fix. I'll close this PR and convert it into a bug report.

PR #3495, opened 2021-10-29
Clarified exception handling from deferred calls in the concurrent programming section.

blais:
I tripped up over the word "directly."


yminsky:
Tried different wording, in #3496

PR #3501, opened 2021-11-06
[gadts] Fix typo 'syntax here is requires' -> 'syntax here requires'

paulcavallaro:
Fix typo 'The syntax here is requires' -> 'The syntax here requires'


yminsky:
Thanks!

PR #3510, opened 2021-12-13
Remove `require-package` labels

Leonidas-from-XIV:
These were only used for `mdx rule` but since the code uses the dune `mdx` stanza now this is declared in the `dune` file instead.

### DIFF
--- a/book/concurrent-programming/README.md
+++ b/book/concurrent-programming/README.md
@@ -989,7 +989,7 @@ let () =
 And that's all we need for a simple but usable definition
 searcher:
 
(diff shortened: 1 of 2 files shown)


NathanReb (approved):
Looks good to me! Thanks!

NathanReb:
A bit of context for @yminsky @avsm: these were leftovers from a time when we used `ocaml-mdx rule` to generate dune rules. I must have forgotten to remove those when upgrading to the dune stanza. They weren't used anymore and removing them has no impact on MDX' behaviour here.

This label, alongside the `ocaml-mdx rule` command will be removed from MDX in the to-be-released-very-soon 2.0.0 version so this helps making RWO compatible.

PR #3515, opened 2022-01-05
pdf build, epub-mobi convert and artifacts in ci

prekel:


prekel:
I note that Pandoc filter isn't working for `epub` - crosslinks are empty. I will try to fix

prekel:
There is no way to pass arguments to pandoc filter for distinguishing `tex` and `epub` with one filter executable. So, I create a filter for epub named `pandoc_rwo_epub` and rename the current filter `pandoc_rwo` to `pandoc_rwo_tex`.

A new filter for epub successfully handles cross-references, but I don't know how to display block-notes in epub. Thus, block-notes are just text index heading-6.

yminsky:
This looks cool! But I'll defer to @avsm for actually reviewing the build rules.

prekel:
The latest pandoc (2.17.1 from brew) breaks the build (https://github.com/realworldocaml/book/runs/5076046684?check_suite_focus=true), so I switch to apt (2.5).

PR #3523, opened 2022-01-17
Put GADTs before Functors.

yminsky:
put the GADT chapter earlier, make a forward-reference to functors in the one place they show up


avsm:
lgtm, as discussed

PR #3538, opened 2022-02-12
Fix pdf hrefs

avsm:


book/functors/README.md
```diff
@@ -1081,24 +1081,20 @@ same basic pattern, including:
...
   sets, and hash heaps.
 
-`Monad.Make`
-: For so-called monadic libraries, like those discussed in Chapters
-  [Error Handling](error-handling.html#error-handling){data-type=xref data-xrefstyle="select:
-  labelnumber"} and
-  [Concurrent Programming With Async](concurrent-programming.html#concurrent-programming-with-async){data-type=xref data-xrefstyle="select:
-  labelnumber"}. Here, the functor is used to provide a collection of
-  standard helper functions based on the `bind` and `return` operators.
+* `Monad.Make` : For so-called monadic libraries, like those discussed in Chapters
```
avsm, line 1092: This switch to a list layout works around a pandoc filter issue

book/json/README.md
```diff
@@ -76,9 +76,10 @@ checking]{.idx}[unit tests]{.idx}
 ##### Installing the Yojson Library
 
 There are several JSON libraries available for OCaml. For this chapter, we've
-picked the [Yojson](https://github.com/mjambon/yojson) library by Martin Jambon.
-It's easiest to install via OPAM by running `opam install yojson`. See
-[the installation instructions](install.html) if you haven't already got opam.
+picked the [Yojson](https://github.com/ocaml-community/yojson) library by Martin Jambon.
```
yminsky, line 79: Maybe we should fix this in a separate PR, but this feels like too much time spent on installing. We opam install things in lots of places. I would just say tell people to install yojson by typing opam install yojson. Like, in "lists and patterns", we write:

> Here, we’ll benchmark these functions using the core_bench library, which can be installed by running opam install core_bench from the command line.

which seems...fine?

yminsky:
Looks good. I have one comment here, but I think it can and should be fixed in a follow-on feature.

PR #3540, opened 2022-02-13
remove inline spans, unnecessary now that we convert to tex as well

avsm:


yminsky:
Why do we have all these skips? Is something not building properly?

Also, it looks like there are lots of span removals in this CR, even though ditch-span is already released? Not sure what's going on there.

PR #3541, opened 2022-02-13
Refresh parsing chapter

avsm:


avsm:
this is layered over #3540

yminsky:
This looks basically fine, but I saw you add some skips, and it wasn't clear to me why they were necessary. That said, still happy to merge.

avsm:
@NathanReb looks like Menhir is broken in opam-monorepo since there's a newer version, but the opam-overlays haven't forward ported my 'check in the manual' fix.

NathanReb:
Ah yeah sorry, failed to get back to this. @samoht mentioned that indeed merlin depends on dune but doens't build properly in a duniverse.

Sorry for the inconvenience, I'll take a look!

CC @Leonidas-from-XIV

Leonidas-from-XIV:
I looked into the issue that the `menhir` binary is not found in the sub-invocation of dune with `--root`, where even adding `(package menhir)` does not work.

It seems like the dependency is not handled correctly (as it works if you run `dune build` before `dune runtest` as it picks up the menhir dependency) so I tried to make a test case for current dune development branch but it seems that is fixed there, so #3551 might actually fix the `menhir` binary issue without any additional work (bar adding the `(package menhir)` stanza again).

avsm:
This is not fixed in dune 3.0.2.

PR #3542, opened 2022-02-13
updates on platform chapter

avsm:


yminsky:
I pushed some more TODOs. I think it's fine to merge as is, but we should address the todos before we're done.

Issue #3547, opened 2022-02-18
Build error: `Undefined control sequence`


sandreenko:
When I try to build there are many warnings and this error that terminates the process:
``` 
Overfull \hbox (48.36967pt too wide) has occurred while \output is active
| [][]| []
[38]
! Undefined control sequence.
<argument> @{} >{\raggedright \arraybackslash 
                                              }p{(\columnwidth - 2\tabcolsep...
l.2883 ...idth - 2\tabcolsep) * \real{0.6351}}@{}}
                                                  
? 
! Emergency stop.
<argument> @{} >{\raggedright \arraybackslash 
                                              }p{(\columnwidth - 2\tabcolsep...
l.2883 ...idth - 2\tabcolsep) * \real{0.6351}}@{}}
                                                  
Output written on book.pdf (44 pages).
Transcript written on book.log.
make: *** [all] Error 1
```
[book.log](https://github.com/realworldocaml/book/files/8099891/book.log)

Does anybody know a fix or could somebody just give me a fresh pdf?

Thank you.

avsm:
This should be fixed now since #3551

PR #3548, opened 2022-02-18
Update dune version in readme.

sandreenko:
The min supported version was bumped several times so now it you install dune2.0.1 you can't run make.

Also, you need to have dozens other packages to successfully compile the book, for example:
 basictex, pandoc and dozen packages like `tlmgr install framed `.

and there are build failures https://github.com/realworldocaml/book/issues/3547

### DIFF
--- a/README.md
+++ b/README.md
@@ -48,7 +48,7 @@ libraries. To set up your RWO development environment you can run:
 
 ```
 opam switch create rwo 4.09.0
-opam install dune=2.0.1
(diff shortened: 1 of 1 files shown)


yminsky:
@avsm Do you know why the "Main workflow" is failing here?

sandreenko:
@yminsky any chance you have a compiled pdf version that you can share? Thanks.

PR #3557, opened 2022-02-21
Add more double-semicolons

Leonidas-from-XIV:
Followup to #3556, the previous one missed out on some references (because I've been fixing these by hand).

Will switch it to ready to review once I have caught up on all.


Leonidas-from-XIV:
There's still some test failures on my machine but this is due to OCaml error messages, the workflow is running on 4.13.1, my switch is 4.12.

avsm:
thanks! I'll merge this and we can fix the CI async

PR #3560, opened 2022-02-21
Add remaining semicolons that were left out before

Leonidas-from-XIV:
For some reason when running the tests I missed out the remaining error messages, so to be sure this time I cleaned the build and built it from scratch.

Really embarassing, that.


avsm:
thanks!

PR #3567, opened 2022-02-25
Use dune 3.0 language + mdx 0.2 stanza 

Leonidas-from-XIV:


avsm:
What's the equivalent of the non-deterministic flag now?

emillon:
This part is removed because it was dead code. `(lang dune 3.0)` adds an implicit empty `mli` file for every executable, which caught this.

As for non-deterministic tests, the setup can be simplified now that both mdx and dune know about the env var (`MDX_RUN_NON_DETERMINISTIC=1 dune runtest` will work so the context workaround is not needed anymore).

avsm:
Thanks! Is the MDX_RUN_NON_DETERMINISTIC documented anywhere? Looking at the MDX README, I can only spot `ocaml-mdx test --non-deterministic` referenced.

Leonidas-from-XIV:
@avsm That's a good point, I added a bit of info about it in https://github.com/realworldocaml/mdx/pull/369. Not extremely comprehensive but it is good to be aware of it.

Leonidas-from-XIV:
One of the issues that is happening is that `core.top` is not loaded. Whenever I try to load it by adding it to the `dune` files, it successfully gets the list of printers, but installing fails for some reason:

```
Unbound value Uuid.pp.
Unbound value Core.Time_ns.pp.
Unbound value Core.Time_ns.Option.pp.
Unbound value Core.Time_ns.Ofday.Option.pp.
Unbound value Core.Time_ns.Ofday.Zoned.pp.
Unbound value Core.Time_ns.Span.Option.pp.
Unbound value Core.Time.pp.
Unbound value Core.Time.Ofday.Zoned.pp.
Unbound value Core.Unix.Cidr.pp.
Unbound value Timezone.pp.
Unbound value Core_kernel.Pid.pp.
Unbound value Core_kernel.Host_and_port.pp.
Unbound value Core_kernel.String_id.pp.
Unbound value Core.Time_ns.Ofday.pp.
Unbound value Core_kernel.Time_ns.Span.pp.
Unbound value Core_kernel.Time.Ofday.pp.
Unbound value Core_kernel.Time.Span.pp.
Unbound value Core_kernel.Date.pp.
Unbound value Base.Float.pp.
Unbound value Base.Int32.pp.
Unbound value Base.Bool.pp.
Unbound value Base.Int63.pp.
Unbound value Base.Int64.pp.
Unbound value Base.Nativeint.pp.
Unbound value Base.Nothing.pp.
Unbound value Base.Sign_or_nan.pp.
Unbound value Base.Int.pp.
Unbound value Base.Sign.pp.
Unbound value Base.Bytes.pp.
Unbound value Base.Char.pp.
Unbound value Base.Uchar.pp.
Unbound value Base.Unit.pp.
Unbound value Base.Error.pp.
Unbound value Base.Info.pp.
Unbound value Base.Exn.pp.
Unbound value Base.Sexp.pp_hum.
```

That's very odd, given that when I try to access values like `Base.Int.pp` in the `.md` files they are successfully found and the proper signature is output.

PR #3569, opened 2022-02-26
Add the .gitignored files from Menhir

Leonidas-from-XIV:
Menhir has a really odd setup

1. The manual is not built with `dune`, it uses `make`
2. The release tarballs contain the files that `make` (or rather LaTeX) generates
3. The git repo does not and in fact `.gitignores` them

So when `menhir` is pulled via `opam-monorepo` it pulls the tarball including all the (generated) files. But when it was being added to this repository `git add` excluded the generated files in `duniverse/menhir/doc/.gitignore`, so when checking out the repo via git and not using `opam-monorepo pull` afterwards would yield a broken `menhir` as [described by Anil](https://github.com/realworldocaml/book/pull/3565#issuecomment-1051867667).

This PR adds the .gitignored files. `gitignore` in `duniverse/` should probably not be respected because it might ignore important files.


avsm:
excellent detective work, thanks! merged a conflict-free version in 72738b5e146ea8da4f5df7803dbfbaec1dd42f85

PR #3571, opened 2022-03-06
Review of files-modules-and-programs

yminsky:
Result of final reread of files-modules-and-programs.

- Some small whitespace and formatting fixes
- Very small language changes
- Explained dune-project file, which is now required by dune.
- Added some cross-references to the platform chapter
- Fixed a bunch of overlong code blocks with ocamlformat. Added a shared ocamlformat for all examples in the chapter, and symlinked .ocamlformat files of individual examples up to that. Maybe that should be done in other chapters...

### DIFF
--- a/book/files-modules-and-programs/README.md
+++ b/book/files-modules-and-programs/README.md
@@ -671,26 +671,28 @@ Error: Unbound value foo
 - : int = 3
 ```
 
-`open` is essential when you want to modify your environment for a standard
-library like `Base`, but it's generally good style to keep the opening of
-modules to a minimum. Opening a module is basically a trade-off between
-terseness and explicitness—the more modules you open, the fewer module
-qualifications you need, and the harder it is to look at an identifier and
-figure out where it comes from.
-
-Here's some general advice on how to deal with `open`s:
+Here's some general advice on how to use `open` effectively.
 
 ### Open modules rarely
 
-Opening modules at the toplevel of a module should be done quite
-sparingly, and generally only with modules that have been specifically
-designed to be opened, like `Base` or `Option.Monad_infix`.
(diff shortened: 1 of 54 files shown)


book/files-modules-and-programs/README.md
```diff
@@ -648,14 +678,19 @@ terseness and explicitness—the more modules you open, the fewer module
...
+Here's some general advice on how to deal with `open`s:
 
-- Opening modules at the toplevel of a module should be done quite sparingly,
-  and generally only with modules that have been specifically designed to be
-  opened, like `Base` or `Option.Monad_infix`.
+### Open modules rarely
 
-- If you do need to do an open, it's better to do a *local open*. There are
-  two syntaxes for local opens. For example, you can write:
+Opening modules at the toplevel of a module should be done quite
```
avsm, line 685: The term 'toplevel' here might be slightly confusing, given it's also the term for the interactive REPL.

book/files-modules-and-programs/README.md
```diff
@@ -648,14 +678,19 @@ terseness and explicitness—the more modules you open, the fewer module
...
+sparingly, and generally only with modules that have been specifically
+designed to be opened, like `Base` or `Option.Monad_infix`.
+
+### Prefer local opens
+
+If you do need to do an open, it's better to do a *local open*. There
```
avsm, line 691:
```suggestion
If you do need to do an `open`, it's better to do a *local open*. There
```

Just a nit: you use `open`s below but a bare "open" here.

book/files-modules-and-programs/README.md
yminsky: Good point. Fixed that with some slightly wider rewording.

PR #3608, opened 2022-04-25
Chapter 'Classes': simplify code

benjub:
One could also remove the extraneous parentheses around `Float.of_int radius`.

I get an error when running this code. This is because Base/Core removes the polymorphic nature of `<=`. Not sure how to best fix it, if one wants to follow the spirit of that example. Maybe
```
  method private contains x' y' =
    let dx = x' - x in
    let dy = y' - y in
    let dist = Float.sqrt (Float.of_int (dx * dx + dy * dy)) in
      Float.(dist <= of_int radius)
``` 
But why not avoid `Float` altogether ?
```
  method private contains x' y' =
    let dx = x' - x in
    let dy = y' - y in
      dx * dx + dy * dy <= radius * radius
```
@yminsky : tell me if that's ok and I'll gladly implement that latter solution in both instances.

### DIFF
--- a/book/classes/README.md
+++ b/book/classes/README.md
@@ -956,8 +956,7 @@ class circle r x y = object(self)
   method private contains x' y' =
     let dx = x' - x in
     let dy = y' - y in
-    let dist = sqrt (Float.of_int (dx * dx + dy * dy)) in
-      dist <= (Float.of_int radius)
+      dx * dx + dy * dy <= radius * radius
 
   method on_click ?start ?stop f =
     on_click ?start ?stop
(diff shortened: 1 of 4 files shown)


yminsky:
This needs a run of "dune runtest; dune promote" to get the README to match the source.

benjub:
This (dune runtest && dune promote) also makes unrelated changes...

Is my int-only solution above ok with you ?

yminsky:
These unrelated changes are not good, and CI still fails. I think you're getting those other promotions because you're not running the right version of OCaml (which is 4.13.1, I believe.)

Maybe I'll try my own version of this simplification.

PR #3610, opened 2022-04-26
Review index

yminsky:
Made a bunch of small tweaks to the index. Note that /'s are meaningful, so when we had an index topic of I/O, that treated I as the heading and O as the subheading! So I got rid of a bunch of those, in addition to a bunch of other miscellaneous fixes.


book/concurrent-programming/README.md
```diff
@@ -443,7 +443,7 @@ example]{.idx}
 The first step is to create a function that can copy data from an input to an
 output. Here, we'll use Async's `Reader` and `Writer` modules, which provide
 a convenient abstraction for working with input and output channels: [Writer
-module]{.idx}[Reader module]{.idx}[I/O (input/output) operations/copying
```
avsm, line 446: good catch! :)

PR #3614, opened 2022-05-04
Fix indenting in pdf

avsm:


avsm:
@yminsky was the indentation in d20b9c1d42febcee9d3e8c859c7e0383364df4a8 intentional, or is the fix right?

yminsky:
the fix is correct!

PR #3615, opened 2022-05-06
menhir chapter: an object is list of (k,v) tuples

sidkshatriya:
menhir chapter: an object is list of (k,v) tuples


sidkshatriya:
Nevermind. `;` separates new values in an OCaml list literal. OCaml is smart enough to make a tuple with a comma. An enclosing `( )` is not required.

PR #3628, opened 2022-05-16
Try newer `num` port

Leonidas-from-XIV:
This PR is just to see if a newer version of the `num` port fixes the build issue by running it on OCaml CI.

I can't reproduce the failure on my own machine, but it did fail on OCaml CI so if it works here then there's a good chance that it fixes the issue.


Leonidas-from-XIV:
This seems to be working, so I am submitting the change [as a PR](https://github.com/dune-universe/num/pull/3) to our `num` port.

PR #3646, opened 2022-08-20
GADT Chapter: Permit the naive variant implementation to compare bools

TyOverby:
The "desired module type" has this signature for `eq`

```ocaml
val if_ : bool t -> 'a t -> 'a t -> 'a t
```

But the naive implementation throws an exception if it receives anything other than two ints.  This isn't really necessary, so I propose a change to the program which allows it.

### DIFF
--- a/book/gadts/README.md
+++ b/book/gadts/README.md
@@ -82,7 +82,7 @@ With that in hand, we can write the evaluator itself.
        | Int _ -> raise Ill_typed)
     | Eq (x, y) ->
       (match eval x, eval y with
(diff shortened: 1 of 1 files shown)


TyOverby:
After getting to the "implementation with GADTs" section and seeing that it also requires the arguments to be ints, maybe the correct fix is to update the module signature instead of the implementations

TyOverby:
never mind!; looks like the "mistake" that I caught was intentional, and used to demonstrate an issue with phantom types...  Serves me right for not finishing the chapter haha

yminsky:
Hah! Glad you're reading closely enough to notice...

PR #3664, opened 2023-05-01
fix bad link to dune-build-info library

jamessouth:


jamessouth:
I would also like to note that the last [link](https://ocamllabs.github.io/ocaml-ctypes) in the final sentence of Chapter 22, Foreign Function Interface, is a 404.

PR #3667, opened 2023-06-25
Mention one needs first to import "base" before opening it

ulugbekna:
It was recently a source of confusion for a new comer reading RWO on his stream - https://www.twitch.tv/theprimeagen/


laksith19:
Hmm yea this line surely can cause some confusion as technically Base should already be opened by default if you create the `~/.ocamlinit`  as per the [install instructions](https://dev.realworldocaml.org/install.html). Base is available after `#require core.top`.

